Issue #4, opened 2015-03-18
Add client port to use in multiline filter stream identity


doubret:
In my case i have several tcp connections from the same host on a logstash tcp input.
When i use the multiline filter, i need to setup a different stream identity for each connection so that multiline can recompose the messages correctly.
For this i needed to modify the tcp.rb code to expose the port of the client connection (only host was exposed).

Would it be possible to make this available in the official plugin ?
Thanks.

jordansissel:
Could you use the multiline codec instead? It doesn't need stream identity because each stream runs with its own codec, and the multiline codec exists to help solve this kind of thing.

doubret:
Good point, i will look into it !
The only thing i need to check is that it works with  charset => "CP1252".
I use NxLog as a log forwarder and then  charset => "CP1252" is important.
I will let you know if the multiline code worked.

Anyway, it makes sense to expose the port. In fact i listen on 3 tcp ports, on each port the format of incoming messages is the same, each port collects events from different type of sources but messages share the same format. It made sense to use the multiline filter because it saves from writing the multiline codec three times...

Thank you.

jordansissel:
@doubret for now, you can work around this by using `add_field`

```
input {
  tcp {
    port => 1234
    add_field => { "port" => 1234 }
  }
}
```

jordansissel:
> It made sense to use the multiline filter because it saves from writing the multiline codec three times...

The multiline codec will perform better because it is executed in the same thread(s) as the tcp input and avoids problems of out-of-order-messages that you'll get if you run multiple filter workers (`bin/logstash -w 4` for example).

doubret:
Nope, i need the client port, not the server port ;)

jordansissel:
Ahh, gotcha. I think we used to have this, but removed it from the 'host' field. Do you want the client port only? or the address and port?

doubret:
When you suggested the codec thing i immediately thought about the thread advantage... I admit it's very temptating as i cannot increase number of workers with multiline filter.

I think i'll go for it.
But the idea of exposing client port is not a bad one i guess.

jordansissel:
sweet. Can you file a separate ticket for the 'client port' feature request? Should be easy to fix. :)

doubret:
adress and port should be good, anything that can identify the stream in the filter, it could be an internal identifier but the simplest is to use the client_host and client_port, it is always unique.

[...]

PR #5, opened 2015-03-19
Expose client port outside input

doubret:
Expose client port outside input so that it can be used in the multiline
filter stream identity for exemple.

See this discussion for more info : https://github.com/logstash-plugins/logstash-input-tcp/issues/4


doubret:
Any update ?

dtrv:
The patch works for me. Several local applications putting multiline-JSON simultaneously to tcp-input.

purbon:
Merged manually in https://github.com/logstash-plugins/logstash-input-tcp/commit/6558ef149c3d774ef6bed2c890e42ff9bd339be1

PR #6, opened 2015-03-31
refactor input specs and shutdown sequence

colinsurprenant:
this also requires the new logstash-core 1.5.0.rc3 and logstash-devutils 0.0.12

### DIFF
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -7,7 +7,7 @@ require "logstash/inputs/tcp"
 require 'stud/try'
 
 describe LogStash::Inputs::Tcp do
-  
+
 
   context "codec (PR #1372)" do
     it "switches from plain to line" do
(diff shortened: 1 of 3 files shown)


spec/inputs/tcp_spec.rb
```diff
@@ -23,13 +24,14 @@
       plugin = LogStash::Inputs::Tcp.new("codec" => LogStash::Codecs::JSON.new, "port" => 0)
       plugin.register
       insist { plugin.codec }.is_a?(LogStash::Codecs::JSONLines)
+      plugin.teardown
     end
   end
 
-  describe "read plain with unicode", :socket => true do
+  it "should read plain with unicode" do
```
purbon: I tend to avoid to use of should in test example naming... a good explanation of why http://betterspecs.org/#should

spec/inputs/tcp_spec.rb
```diff
@@ -23,13 +24,14 @@
       plugin = LogStash::Inputs::Tcp.new("codec" => LogStash::Codecs::JSON.new, "port" => 0)
       plugin.register
       insist { plugin.codec }.is_a?(LogStash::Codecs::JSONLines)
+      plugin.teardown
     end
   end
 
-  describe "read plain with unicode", :socket => true do
```
purbon: With the describe clean, you got rid of the socket filter, is this something we want to do? notice, I'm happy to get rid of most of this filters :-)

colinsurprenant: yes, the :socket flag was only there because the socket related tests were broken.

PR #8, opened 2015-05-10
[WIP] refactor plugin shutdown

colinsurprenant:
refactor for the new shutdown semantic.

note that the specs were not refactored as a whole be new specs were added, using the new style, based on the udp input specs. some TODOs have been commented to note some duplicate code and DRYing that we could do in a followup issue/PR.

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -1,6 +1,5 @@
 # encoding: utf-8
 require "logstash/inputs/base"
-require "logstash/namespace"
 require "logstash/util/socket_peer"
 
 # Read events over a TCP socket.
@@ -22,11 +21,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   # When mode is `client`, the port to connect to.
   config :port, :validate => :number, :required => true
 
-  # The 'read' timeout in seconds. If a particular tcp connection is idle for
-  # more than this timeout period, we will assume it is dead and close it.
-  #
-  # If you never want to timeout, use -1.
-  config :data_timeout, :validate => :number, :default => -1
+  config :data_timeout, :validate => :number, :default => -1, :deprecated => "This setting is not used by this plugin. It will be removed soon."
 
   # Mode to operate in. `server` listens for client connections,
   # `client` connects to a server.
@@ -53,12 +48,14 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
 
   def initialize(*args)
     super(*args)
-    @interrupted = false
+
+    # threadsafe socket bookkeeping
     @server_socket = nil
     @client_socket = nil
-  end # def initialize
+    @connection_sockets = {}
+    @socket_mutex = Mutex.new
+  end
 
-  public
   def register
     require "socket"
     require "timeout"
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/tcp.rb
```diff
@@ -86,48 +81,114 @@ def register
...
+  def run(output_queue)
+    if server?
+      run_server(output_queue)
+    else
+      run_client(output_queue)
+    end
+  end
+
+  def stop
+    super
```
talevy: not needed anymore, right?

lib/logstash/inputs/tcp.rb
```diff
@@ -22,12 +21,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   # When mode is `client`, the port to connect to.
   config :port, :validate => :number, :required => true
 
-  # The 'read' timeout in seconds. If a particular tcp connection is idle for
-  # more than this timeout period, we will assume it is dead and close it.
-  #
-  # If you never want to timeout, use -1.
-  config :data_timeout, :validate => :number, :default => -1
```
talevy: why is there a removed option? if it is unused, shouldn't we set it as `deprecated`?

purbon: agreed, should probably be set to deprecated, specially in this case where it says ```If a particular tcp connection is idle```, I see people still using it.

what do you think?

colinsurprenant: reverted

lib/logstash/inputs/tcp.rb
```diff
@@ -86,48 +83,121 @@ def register
...
+    end
+  ensure
+    server_socket.close rescue nil
+  end
+
+  def run_client(output_queue)
+    while !stop?
+      set_client_socket(TCPSocket.new(@host, @port))
+
+      if @ssl_enable
```
andrewvc: Have you considered extracting this block into a separate method?

lib/logstash/inputs/tcp.rb
```diff
@@ -86,48 +83,121 @@ def register
         @ssl_context.cert_store = @cert_store
         @ssl_context.verify_mode = OpenSSL::SSL::VERIFY_PEER|OpenSSL::SSL::VERIFY_FAIL_IF_NO_PEER_CERT
       end
-    end # @ssl_enable
+    end
+
+    # note that since we are opening a socket in register, we must also make sure we close it
```
andrewvc: This makes me ask the question, why is the socket opened in register vs. run?

lib/logstash/inputs/tcp.rb
```diff
@@ -86,48 +83,121 @@ def register
...
+    else
+      run_client(output_queue)
     end
-  end # def register
+  end
+
+  def stop
+    # force close all sockets which will escape any blocking read with a IO exception
+    # and any thread using them will exit
+    server_socket.close rescue nil
```
andrewvc: I'm always wary of rescue nil. Can we rescue a specific class?

andrewvc: It'd be preferable to have a function like

```ruby
def safe_close_socket(socket)
  client.socket.close
rescue IOExceptionOrWhatever => e
  #noop
rescue StandardError => e
  @logger.error("An unexpected error happened while closing a socket!", :message => e.message)
end
```

lib/logstash/inputs/tcp.rb
```diff
@@ -86,48 +83,121 @@ def register
...
+        begin
+          client_socket.connect
+        rescue OpenSSL::SSL::SSLError => e
+          @logger.error("SSL Error", :exception => e, :backtrace => e.backtrace)
+          sleep(1) # prevent hammering peer
+          next
+        rescue
+          # if this exception occured while the plugin is stopping
+          # just ignore and exit
+          raise unless stop?
```
andrewvc: Good job watching out for corner cases!

lib/logstash/inputs/tcp.rb
colinsurprenant: opening it in register makes testing easier, we can create a client socket before running the plugin.

lib/logstash/inputs/tcp.rb
colinsurprenant: we don't care about potential close exception and it also makes not having to test for invalid, already closed or nil sockets. catch all.

lib/logstash/inputs/tcp.rb
colinsurprenant: It could I guess, and a lot of code could be better refactored here too. tried to keep it to minimal.

andrewvc:
Are these specs passing for you @colinsurprenant ? I'm currently seeing them hang indefinitely.

The code looks alright otherwise.

lib/logstash/inputs/tcp.rb
colinsurprenant: but that being say, using `foo rescue nil` should generally speaking be avoided :P

colinsurprenant:
is it running agaist the lastest core code base? I've run the tests in a loop to catch potential timing issues and it always passed... weird.

lib/logstash/inputs/tcp.rb
jordansissel: @colinsurprenant The way I read it, your intent is to say "Close unless it's already closed, and then I don't really care about the error" - in this case, I'm ok with `foo.bar rescue nil` to just avoid some mess of exception catching that we're going to ignore anyway. +1 on using 'rescue nil' here since it makes (to me) it clear the intent. Maybe a comment? "Close, but if close fails (already closed?) we don't care" or something

spec/inputs/tcp_spec.rb
```diff
@@ -214,35 +212,67 @@
...
       end
 
-      client_threads = inputs[0].instance_variable_get("@client_threads")
+      before(:each) do
+        subject.register
+      end
 
-      # close all sockets and make sure there is not more pending threads
-      sockets.each{|socket| socket.close}
+      it "should receive events been generated" do
```
andrewvc: This spec is where it hangs for me

andrewvc:
@colinsurprenant It hangs for me here https://github.com/logstash-plugins/logstash-input-tcp/pull/8/files#r40319932

This is running of `2.0.0.snapshot2`. Is there another ref I should be basing this off of?

andrewvc:
Running with --debug I get it repeating the INFLIGHT EVENTS REPORT message

```
Accepted connection {:client=>"127.0.0.1:62233", :server=>"0.0.0.0:48607", :level=>:debug, :file=>"logstash/inputs/tcp.rb", :line=>"177", :method=>"block in server_connection_thread"}
config LogStash::Codecs::Line/@charset = "UTF-8" {:level=>:debug, :file=>"logstash/config/mixin.rb", :line=>"111", :method=>"block in config_init"}
stopping {:plugin=><LogStash::Inputs::Tcp port=>48607, debug=>false, codec=><LogStash::Codecs::Line charset=>"UTF-8">, host=>"0.0.0.0", data_timeout=>-1, mode=>"server", ssl_enable=>false, ssl_verify=>false, ssl_key_passphrase=><password>>, :level=>:debug, :file=>"logstash/inputs/base.rb", :line=>"115", :method=>"do_stop"}
["INFLIGHT_EVENTS_REPORT", "2015-09-24T08:45:33-05:00", {"input_to_filter"=>0, "filter_to_output"=>0, "outputs"=>[]}] {:level=>:warn, :file=>"logstash/util/reporter.rb", :line=>"26", :method=>"report"}
["INFLIGHT_EVENTS_REPORT", "2015-09-24T08:45:33-05:00", {"input_to_filter"=>0, "filter_to_output"=>0, "outputs"=>[]}] {:level=>:warn, :file=>"logstash/util/reporter.rb", :line=>"26", :method=>"report"}
```

That's where it hangs

andrewvc:
@colinsurprenant AHA. I was running the specs against JRuby 9.0.0.0 . It worked when I switched to 1.7.20 .

In that case LGTM, but that does not bode well for the future. Any ideas why?

colinsurprenant:
@andrewvc lots have changed IO wise in 9k, we'll have to investigate. Note that we should certainly test against 9k and report problems in JRuby if we find them. OTOH 9k is definitely not mature enough to consider it for a release soon.

jsvd:
LGTM, tests passed, tested manually server/client without ssl, everything worked.

lib/logstash/inputs/tcp.rb
colinsurprenant: +1 comment

colinsurprenant:
squashed, merging.

Issue #9, opened 2015-05-17
tcp input SSL chain/CA configuration issues


jordansissel:
(This issue was originally filed by @MikeN123 at https://github.com/elastic/logstash/issues/1956)

---

At the moment, the SSL configuration options of the tcp input are fairly limited, and some are even useless. The following issues exist:

* It is not possible to specify `extra_chain_cert` for the server certificate, thus if a certificate chain is necessary this can only come from the CA store.
* The CA store (system-store or specified file) is only activated if `ssl_verify = true` (requiring client verification), so without client verification it is not possible to send a proper certificate chain.
* The system-store is always included if `ssl_verify = true`, making client certificate validation useless, since a client can send any certificate from any recognised CA, for any name.

I propose the following changes:

* Allow `extra_chain_cert` to be set, to specify a certificate chain (extra property)
* Allow the certificate store to be initialised from the system store even when not using client certificates (extra property: `ssl_set_default_paths` or something like that, and moving a couple of lines of code)
* Do not use the system store for client certs by default (i.e., make the previous property `false` by default). This may break existing configurations (but in these cases the peer verification was probably useless anyway).

purbon:
Fixed most of this issues in https://github.com/logstash-plugins/logstash-input-tcp/pull/28 

@MikeN123

MikeN123:
Sorry, but looking at the PR, the main issue that made us switch is still present. If client SSL verification is on (`ssl_verify = true`), it will load the system's default store (`cert_store.set_default_paths`). This makes client certificate validation totally useless, as I can send any certificate I want, as long as it is signed by a CA.

purbon:
@MikeN123 very interesting situation, I wonder about pros and cons of removing the system CA. If I understand you properly, what you just say is you want to "authorize" your clients by only allowing clients with certificates that are being issued by your own CA, right?

purbon:
@MikeN123 at my understanding certificate validation is not useless like is right now, it's doing it job if we trust CA to do a proper work. Would really love to know more about your security concerns here. thanks a lot.

MikeN123:
It is useless, unless you manually verify the DN/subject of each client certificate. For that you would have to define a filter or something like that in logstash. I simply do not want unauthorized users to connect, and that is impossible with the current implementation. I also do not want to use a public CA, as that's a rather costly solutions (and I don't think anyone else does use a public CA for client certificates, except for things like e-mail where certificates are more about identification than authorization).

[...]

PR #14, opened 2015-07-23
Extend tcp input

gbu-censhare:
Hi,

i exposed some of the OpenSSL Parameters in the Configuration

a) the CA-Certificate for the server might not be also trusted for clients
b) you might not trust the system ca path for your client certificates
c) you might define the verify depth
d) you might want to traverse the directory and manually add all files to the truststore (we want to add self-signed certifiates for each client host)

I'm not sure, if my naming pattern is okay with yours, Changes should be backwards compatibile.

Thanks,

Georg

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -38,12 +38,27 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   # Verify the identity of the other end of the SSL connection against the CA.
   # For input, sets the field `sslsubject` to that of the client certificate.
   config :ssl_verify, :validate => :boolean, :default => false
+  
+  # Verify the identity of the other end of the SSL connection against the CA.
+  # depth of ca-certificates to be checked
+  config :ssl_verify_depth, :validate => :number, :default => -1
 
   # The SSL CA certificate, chainfile or CA path. The system CA path is automatically included.
+  # Used for the Client Trust Store
   config :ssl_cacert, :validate => :path
+
+  # Include system CA path to Client Trust Store
+  config :ssl_client_ca_include_system_store, :validate => :boolean, :default => true
+
+  # Instead of including every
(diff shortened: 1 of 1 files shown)


MikeN123:
Seems related or fixing #9 :-)

harbulot:
I've just made a rather long comment on issue #9 regarding the trusted CA settings.

It looks like this patch is about similar concepts, but there are still a number of issues. (In particular, I was suggesting using `ca_file` and `ca_path`, in the same way as more OpenSSL-based tools do.)

* Adding `ssl_verify_depth` is a good idea. It's not always used, but it's a reasonable option to have.
* `ssl_client_ca_include_system_store`: including the system store by default rarely make sense as a server anyway. If it's done, it should indeed be done via an option (preferably default to `false`), and not always done (as it has been the case since PR #28 unfortunately).
* `ssl_client_ca_iterate_files` sounds like it can be useful. At least it wouldn't force to follow the hash naming conventions of OpenSSL. That said, since `ssl_cacert` is deprecated nowadays anyway, we'd need another option to tell which directory to iterate through (and `ca_path` would already have an OpenSSL-related expectation, if implemented). Maybe having a single option that's a path to a directory (let's say `ssl_ca_iterated_path`): only used and iterated through if it's set.
* It's a good idea to be able to specify the certificate chain, but that's not what this patch does with its `ssl_cert_cachain` option (documented as "*The SSL CA certificate, chainfile or CA path.*"). Setting `ssl_context.ca_file` or `ssl_context.ca_cert` is actually very different from configuring the (local) certificate chain (which is done via `ssl_context.extra_chain_cert`).
* I would also avoid calling anything `ssl_client_ca_*`. This can be confusing. There already is something confusingly called `ssl_context.client_ca`: to simplify, it's what's used to indicate the list of CAs in the CertificateRequest (implicitly the ones the server is willing to trust, but not necessarily).

kares:
this would need a revamp and since a CLA has been missing for several years I am going to close.

PR #18, opened 2015-09-26
refactor & cleanup plugin structure

colinsurprenant:
this is a post-new-shutdown-semantic in #8 cleanups and refactor. 

@jsvd take a look and let's see if it can inspire/share for the udp input
@ph would like to pick your brain on the socket select + SSL work you did or experiment in the lumberjack input
@andrewvc you suggested some refactors in #8 let me know what you think

This is not yet ready to merge, tests are passing but I haven't done any real-world tests. Let

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -2,6 +2,9 @@
 require "logstash/inputs/base"
 require "logstash/util/socket_peer"
 
+require "socket"
+require "openssl"
+
 # Read events over a TCP socket.
 #
 # Like stdin and file inputs, each event is assumed to be one line of text.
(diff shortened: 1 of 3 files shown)


ph:
@colinsurprenant Concerning the `IO.select` I need to do stress testing to make sure it 'work'  every time, should be able to finish that this week.

colinsurprenant:
@ph perfect, so let's keep that idea for later, once we've identified the correct `IO.select

lib/logstash/inputs/tcp.rb
```diff
@@ -136,6 +110,7 @@ def run_server(output_queue)
         # start a new thread for each connection.
         server_connection_thread(output_queue, socket)
       rescue OpenSSL::SSL::SSLError => e
+        # TODO: (colin) not sure I undestand the logic of logging & ignoring ssl errors here
```
jsvd: my understanding is that if a user connects with an incorrect ssl setup, an exception is raised on the server side (here), but the only thing we care about is dropping this client's connection and accept the next one

lib/logstash/inputs/tcp.rb
```diff
@@ -222,9 +180,72 @@ def read(socket)
...
+
+    @logger.debug? && @logger.debug("Opened connection", :client => "#{socket.peer}")
+
+    socket
+  rescue OpenSSL::SSL::SSLError => e
+    @logger.error("SSL Error", :exception => e, :backtrace => e.backtrace)
+    # catch all rescue nil on close to discard any close errors or invalid socket
+    socket.close rescue nil
+    sleep(1) # prevent hammering peer
+    retry
```
jsvd: in the case of a SSLError on the `mode => client` the plugin should terminate, IMO.

lib/logstash/inputs/tcp.rb
colinsurprenant: ok. make sense so we should probably add an explicit `socket.close rescue nil` here too then

lib/logstash/inputs/tcp.rb
colinsurprenant: hmm, good question, this was the previous logic I reproduced here. I believe the original intent was  to try and connect ad-nauseam hoping the other side would eventually get fixed. 

Also, if the plugin terminates, the pipeline will restart it so in the end it will end up being the same. Any particular reason exiting to the pipeline and restarting would be preferable?

jsvd: oh yes the pipeline would restart the input plugin. this would often be a fatal error and logstash should just terminate.
but I agree, it's out of scope, we can deal with it later

lib/logstash/inputs/tcp.rb
```diff
@@ -2,6 +2,10 @@
 require "logstash/inputs/base"
 require "logstash/util/socket_peer"
 
+require "socket"
+require "timeout"
```
jsvd: timeout doesn't seem to be used anywhere

colinsurprenant: ah good catch. that's old dead require, will clean

jsvd:
minor comments, code LGTM, tests pass :ship:

lib/logstash/inputs/tcp.rb
colinsurprenant: I think in practice we want the input plugins to retry as much as possible when there is such a failure. let's clarify this in terms of input plugin strategy and also document it for input plugin authors. /cc @jordansissel

lib/logstash/inputs/tcp.rb
colinsurprenant: added

jsvd:
@colinsurprenant please also update changelog before merge :)

colinsurprenant:
@jsvd :+1: 

rebasing and merging.

PR #21, opened 2015-10-08
remove usage of RSpec.configure

### LINKED ISSUE #20: Refactor the test helper to not use Rspec.configure.
We have some issues when plugins are redefining the Rspec config, see 
https://github.com/elastic/logstash/issues/3997 The easiest way is to create a simple class to extract that functionality and don't redefine rspec config option in the `spec_helper`.

See @purbon's nice implementation in https://github.com/logstash-plugins/logstash-input-udp/pull/15/files

colinsurprenant:
fixes #20 

also fixed a potential bug, using `nevent` in the spec helper `TcpHelpers` class. It **was** working because `nevent` was defined in the scope, but if another local var was used it would have failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
+## 2.0.3
+ - removed usage of RSpec.configure, see https://github.com/logstash-plugins/logstash-input-tcp/pull/21
+## 2.0.2
+ - refactored & cleaned up plugin structure, see https://github.com/logstash-plugins/logstash-input-tcp/pull/18
 ## 2.0.0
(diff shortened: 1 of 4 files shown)


colinsurprenant:
ready for review @ph @jsvd

colinsurprenant:
since per @ph this can cause rspec failures we should merge this asap.

jsvd:
LGTM, tests run. (also tried and worked well with logstash 2.0 branch)

colinsurprenant:
I just bumped version and added changelog, final-final LGTM?

jsvd:
LGTM :ship: it

PR #28, opened 2015-11-27
Enhance extra chain of certificates and include ton of ssl tests

### LINKED ISSUE #3: Allow to specify SSL certificate chain
If we have a certificate that is signed with intermediate CA that is not in the trust store then clients cannot verify the certificate.

Currently we can only set ssl_cert and ssl_key and this cert is sent to the client for verification. Would it please be possible to use Ruby's extra_chain_cert?

extra_chain_cert:
An Array of extra X509 certificates to be added to the certificate chain.

I'm using syslog-ng to send logs to logstash. On logstash end there is a TCP input:

input {
  tcp {
    port => 51000
    type => syslog

    ssl_enable => true
    ssl_key    => "/etc/ssl/cert.key"
    ssl_cert   => "/etc/ssl/cert.pem"
  }  
}

Syslog-ng:
destination loghost { tcp("host" port(51000) tls(peer-verify("required-trusted")) ); };
log { source(src); destination(loghost); };

However I'm only seeing errors in logstash:
{:timestamp=>"2015-03-05T14:44:10.322000+0000", :message=>"SSL Error", :exception=>#<OpenSSL::SSL::SSLError: Received fatal alert: unknown_ca>, :backtrace=>["org/jruby/ext/openssl/SSLSocket.java:216:in `accept'", "file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/ssl-internal.rb:164:in `accept'", "/opt/logstash/lib/logstash/inputs/tcp.rb:160:in `run_server'", "org/jruby/RubyKernel.java:1521:in `loop'", "/opt/logstash/lib/logstash/inputs/tcp.rb:157:in `run_server'", "/opt/logstash/lib/logstash/inputs/tcp.rb:148:in `run'", "/opt/logstash/lib/logstash/pipeline.rb:163:in `inputworker'", "/opt/logstash/lib/logstash/pipeline.rb:157:in `start_input'"], :level=>:error}

If I check the the sent certificates with openssl s_client:
openssl s_client -host host -port 51000 -showcerts

I only see the first certificate (even though I specified the whole chain in cert.pem).

purbon:
This PR clarify support for adding extra certificates to the X509 verification chain, also add tests for SSL connections using this plugin.

Changes introduced by this PR are:

* Deprecate ssl_cacert as it's confusing, does it job but when willing to add a chain of certificated the name and behaviour is a bit confusing. For now this PR add an extra config that allows you to specify a list of certificates path that will be added to the CAStore.
* Make ssl_verify=true as a default value, if using CERT but not performing validation is not reasonable as security might be compromised in the meanwhile.
* Add test to verify behaviour under different SSL connection circumstances, for example when verification is not performed, when verification is on, but the certificate validation fail, etc.

Fixes #3 and #9


purbon:
@guyboertje would you be so nice to review this changes? thanks a lot.

guyboertje:
Ran tests OK.  LGTM

harbulot:
I've just [commented](https://github.com/logstash-plugins/logstash-input-tcp/issues/33#issuecomment-195085362))on issue #33, but `ssl_extra_chain_certs` actually introduced another kind of confusion unfortunately.

The sever certificate chain and the collection of CA certificates used as trust anchors for client-certificate authentication should really use separate settings (and `ssl_extra_chain_certs` as it is isn't really a chain).

harbulot:
@purbon

When you say: 
> Make ssl_verify=true as a default value, if using CERT but not performing validation is not reasonable as security might be compromised in the meanwhile.

It's true `ssl_verify=true` is better, but we're on the server side here (input). This is forcing client-certificate authentication. Is it a deliberate choice to want to use client certificates by default for all TCP inputs? This seems unusual, although it can make sense for a logging system, without other ways of authenticating the client, indeed.

(Of course, clients should always verify the server certificate, but that's not up to this code, whatever the value of `ssl_verify` is.)

PR #38, opened 2016-01-12
properly convert sslsubject to string & specs

colinsurprenant:
the `Event["sslsubject"]` field was assigned the `OpenSSL::X509::Name` object - `to_s` need to be called on it to assign the `OpenSSL::X509::Name` string representation in the `Event`. 

The problem surfaced when using the logstash-core-event-java where the types of objects assignable to Event fields is stricter. 

I also added specs for this and did some cosmetic cleanups.

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -153,7 +153,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
       codec.decode(read(socket)) do |event|
         event["host"] ||= client_address
         event["port"] ||= client_port
-        event["sslsubject"] ||= socket.peer_cert.subject if @ssl_enable && @ssl_verify
+        event["sslsubject"] ||= socket.peer_cert.subject.to_s if @ssl_enable && @ssl_verify
         decorate(event)
         output_queue << event
       end
@@ -176,7 +176,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
     codec.respond_to?(:flush) && codec.flush do |event|
       event["host"] ||= client_address
       event["port"] ||= client_port
-      event["sslsubject"] ||= socket.peer_cert.subject if @ssl_enable && @ssl_verify
+      event["sslsubject"] ||= socket.peer_cert.subject.to_s if @ssl_enable && @ssl_verify
       decorate(event)
       output_queue << event
     end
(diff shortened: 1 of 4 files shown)


lib/logstash/inputs/tcp.rb
```diff
@@ -153,7 +153,7 @@ def handle_socket(socket, client_address, client_port, output_queue, codec)
       codec.decode(read(socket)) do |event|
         event["host"] ||= client_address
         event["port"] ||= client_port
-        event["sslsubject"] ||= socket.peer_cert.subject if @ssl_enable && @ssl_verify
+        event["sslsubject"] ||= socket.peer_cert.subject.to_s if @ssl_enable && @ssl_verify
```
ph: I understand the problems and how this change fix it, but let me ask a few question.

Oh, So the setter on java side support what kind of objects? `String`, `Number`, `Timestamp`?
Just wondering, It cannot be done transparently without having an implicit conversion on the ruby side?

colinsurprenant: it was always an implicit rule to only have strings, numerics and `Timestamp` objects in the `Event` fields but this is somewhat made explicit in the Java `Event` because of the explicit Java<->Ruby type conversions. If a type is not know it will throw an exception.

ph: Thanks, I am okay with limiting the number of types.

ph:
LGTM! We might catch other similar error in other plugins

lib/logstash/inputs/tcp.rb
colinsurprenant: It's not really optional and/or debatable, in the end it is required to facilitate serialization and the internal Java event implementation cannot use Ruby objects, which do not/can not support the Java serializable interface. 

And, in any way, it would not be a good idea to divert to that implicit rule, having anything else than string and numeric would not make sense anyway.

colinsurprenant:
normally we should not get too many, if we do it is probably because of an oversight like here and is a bug anyway! in the end that will make the code better :D

colinsurprenant:
will bump version and publish, thanks.

ph:
Small related question what was the error raised? `IncorrectType`?
Also how do you run the test for a specific plugin but using `logstash-event-java` and not the ruby one?

colinsurprenant:
The error raised was
```
java.lang.IllegalArgumentException: Missing Ruby class handling for full class name=org.jruby.ext.openssl.X509Name, simple name=X509Name
```

one of the difficulty was that this exception was swallowed in the rescue clauses on the tcp input and running the specs does not show the debug & error logs outputted so it took me some time to figure that this was the problem. 

I will update the README in https://github.com/elastic/logstash/tree/master/logstash-core-event-java to include instruction for the plugins specs.

ph:
:heart: awesome.

PR #43, opened 2016-01-29
Improving tls stability

jordansissel:
This fixes two bugs in the TCP input.

1) When an error occurs in accepting a new socket, common with TLS scenarios, the _previous_ connection accepted would accidentally be closed. Oops!

2) When providing logging during an exception, we were trying to log the socket's peer address. However, asking a socket for it's peer address _after_ the socket has been closed will cause an exception, so our exception handler was throwing an exception when we tried to log. Oops!

I've added some additional tests to help cover certain kinds of scenarios.

I wanted to write a regression test to verify that the two above scenarios no longer occurred, but I am not sure of the complexity yet to write such a test (with multiple active connections, injecting one or two bad/corrupt connections and verifying the result). Since I don't want to delay the merging or review of this change, I'll submit now. I have manually verified that Logstash doesn't crash under the two above scenarios and it behaves, what I believe to be, correctly (the bad connection is dropped and existing connections are unharmed).

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -115,7 +115,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
       rescue OpenSSL::SSL::SSLError => e
         # log error, close socket, accept next connection
         @logger.error("SSL Error", :exception => e, :backtrace => e.backtrace)
-        socket.close rescue nil
       rescue => e
         # if this exception occured while the plugin is stopping
         # just ignore and exit
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -149,6 +149,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   end
 
   def handle_socket(socket, client_address, client_port, output_queue, codec)
+    peer = "#{client_address}:#{client_port}"
     while !stop?
       codec.decode(read(socket)) do |event|
         event["host"] ||= client_address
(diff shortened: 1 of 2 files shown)


ph:
I will review it, Actually I went through some of the similar issue when I did the refactoring of the beats input.

spec/inputs/tcp_spec.rb
```diff
@@ -5,11 +5,14 @@
...
 require "logstash/inputs/tcp"
 require "stud/try"
+require "stud/task"
 require "flores/pki"
 require "openssl"
 
 require_relative "../spec_helper"
 
+#Cabin::Channel.get(LogStash).subscribe(STDOUT)
+#Cabin::Channel.get(LogStash).level = :debug
```
ph: Remove it maybe?

jordansissel: +1

spec/inputs/tcp_spec.rb
```diff
@@ -236,229 +239,139 @@
...
 
-            it "should contain sslsubject" do
-              events.each do |event|
-                expect(event["sslsubject"]).to eq("/DC=org/DC=ruby-lang/CN=RubyC_Cert")
-              end
-            end
-          end
+        context "with client certificate problems" do
+          context "using an expired certificate"
+          context "using an untrusted certificate"
```
ph: I am okay to have implementation pending test.

lib/logstash/inputs/tcp.rb
```diff
@@ -159,16 +159,16 @@ def handle_socket(socket, client_address, client_port, output_queue, codec)
       end
     end
   rescue EOFError
-    @logger.debug? && @logger.debug("Connection closed", :client => socket.peer)
+    @logger.debug? && @logger.debug("Connection closed", :client => peer)
   rescue Errno::ECONNRESET
-    @logger.debug? && @logger.debug("Connection reset by peer", :client => socket.peer)
+    @logger.debug? && @logger.debug("Connection reset by peer", :client => peer)
```
ph: why peeer! Oh whyyyy! Actually I do the same hack in the `Connection` class in beats.

lib/logstash/inputs/tcp.rb
```diff
@@ -115,7 +115,6 @@ def run_server(output_queue)
       rescue OpenSSL::SSL::SSLError => e
         # log error, close socket, accept next connection
         @logger.error("SSL Error", :exception => e, :backtrace => e.backtrace)
-        socket.close rescue nil
```
ph: I am wondering if we dont have this issue in beats input I will check that.

jordansissel: I comfort myself poorly with this kind of error because any language with "unassigned value"-kind of detection at compile-time would have caught this error :|

ph:
Change LGTM, test seems to cover the crash scenario and the peer fixes will certainely fix the problem since I have experienced the same behavior in the beat refactor. I still find it weird that the peer return an IOError, I would have assumed that this information is cached.

PR #53, opened 2016-06-14
Change log level for OpenSSL error

ph:
SSL Error can happen when a handshake is not correctly completed like
when a LB is in front of the tcp input and does keep alive to check if
the port is still up and running. We should not mark them as error but
instead as debug.


jsvd:
LGTM

chandlermelton:
Yes please. Thanks!

PR #56, opened 2016-07-20
Add metrics to tcp input

purbon:
Add metrics to track:
- # events processed.
- # connections open.
- errors based on SSL or connection resets.

related to https://github.com/elastic/logstash/issues/5668 , https://github.com/elastic/logstash/issues/5607

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -171,7 +171,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
       end
     end
   rescue EOFError
-    @metric_errors.increment(:connection)
     @logger.debug? && @logger.debug("Connection closed", :client => peer)
   rescue Errno::ECONNRESET
     @metric_errors.increment(:connection)
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -81,6 +81,8 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
     # a situation where register is called but not run & stop.
 
     self.server_socket = new_server_socket if server?
+
+    @metric_errors = metric.namespace(:errors)
   end
 
   def run(output_queue)
@@ -115,9 +117,11 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
       begin
         socket = add_connection_socket(server_socket.accept)
         # start a new thread for each connection.
+        metric.increment(:connections)
         server_connection_thread(output_queue, socket)
       rescue OpenSSL::SSL::SSLError => e
         # log error, close socket, accept next connection
+        @metric_errors.increment(:ssl)
         @logger.debug? && @logger.debug("SSL Error", :exception => e, :backtrace => e.backtrace)
       rescue => e
         # if this exception occured while the plugin is stopping
@@ -133,6 +137,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   def run_client(output_queue)
     while !stop?
       self.client_socket = new_client_socket
+      metric.increment(:connections)
       handle_socket(client_socket, client_socket.peeraddr[3], client_socket.peeraddr[1], output_queue, @codec.clone)
     end
   ensure
@@ -146,6 +151,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
         @logger.debug? && @logger.debug("Accepted connection", :client => s.peer, :server => "#{@host}:#{@port}")
         handle_socket(s, s.peeraddr[3], s.peeraddr[1], q, @codec.clone)
       ensure
+        metric.decrement(:connections)
         delete_connection_socket(s)
       end
     end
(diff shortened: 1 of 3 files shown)


purbon:
@jordansissel @ph your thoughts here are very welcome.

purbon:
seems the metrics object is not available, might that be non released snapshot? because I tested this with master code and works properly.

jordansissel:
Oooh exciting! I will review shortly. <3

jordansissel:
I think I found a bug (maybe?) in the metrics api. Metrics from this PR show up like this:

```
    "pipeline" : {
      "inputs" : [ {
        "id" : "events",
        "events" : 762
      }, {
        "id" : "connections",
        "connections" : 0
      }, {
        "id" : "errors",
        "connection" : 763
      } ],
```

I think it should be nested underneath a "tcp" section under "inputs" somehow. I'm not sure yet how this happens.

lib/logstash/inputs/tcp.rb
```diff
@@ -164,11 +171,14 @@ def handle_socket(socket, client_address, client_port, output_queue, codec)
       end
     end
   rescue EOFError
+    @metric_errors.increment(:connection)
```
jordansissel: Should we consider EOF an error? EOF feels more like "The connection is done" and not really an error.

jordansissel:
All the tests seem to have failed because something about metrics is not setup during rspec: https://travis-ci.org/logstash-plugins/logstash-input-tcp/builds/146147210

I think this is probably a bug in our specs somewhere.

acchen97:
@purbon thanks. We'll need to make sure this stuff get documented as well in the monitoring API docs. /cc @suyograo

lib/logstash/inputs/tcp.rb
ph: Agree, `EOF` should be tracked as a connection error, same thing with SSLError. 
Theses kinds of errors are so common in a **normal** lifecycle of an application and tracking them will give probably result in new issues created in the tracker.

I remember one time, when I decided to log theses error in the beat plugin with the `error` level, people using LB in front of Logstash were creating issues asking *what is going on*, only to discover that the keep alive check were causing them ;)

ph:
Good job @purbon I am still flabbergasted at how easy it is to add them :)

ph:
Concerning the test, this plugin use the old helpers to setup a pipeline we might have a bug in them.

purbon:
catching up on this, sorry being offline for a few days.

purbon:
Interesting, just run the test locally, again, as I did before pushing this PR a while ago and the test pass. See:

```
skywalker% bundle exec rspec
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}

Randomized with seed 56981
.............

Finished in 10.34 seconds (files took 6.29 seconds to load)
13 examples, 0 failures

Randomized with seed 56981

skywalker% git status
On branch feature/metrics
Your branch is up-to-date with 'origin/feature/metrics'.

nothing to commit, working directory clean
skywalker% git log
commit a5c9f8fae430e1820c43ab7aa4d5e232e0b82b85
Author: Pere Urbon-Bayes <pere.urbon@elasticsearch.com>
Date:   Wed Jul 20 17:59:44 2016 +0200

    update changelog
```

might be a cache issue in travis?  checking.

purbon:
and now build is green without any change, might be cache, might be unreleased snapshot 🐼 !

purbon:
also clean up the EOFError metrics as agreed, I guess this is now ok for another review. what do you think?

lib/logstash/inputs/tcp.rb
```diff
@@ -146,6 +151,7 @@ def server_connection_thread(output_queue, socket)
         @logger.debug? && @logger.debug("Accepted connection", :client => s.peer, :server => "#{@host}:#{@port}")
         handle_socket(s, s.peeraddr[3], s.peeraddr[1], q, @codec.clone)
       ensure
+        metric.decrement(:connections)
```
ph: I am wondering if it would make testing easier if we move the metric decrement in the `#delete_connection_socket` method.

lib/logstash/inputs/tcp.rb
```diff
@@ -133,6 +137,7 @@ def run_server(output_queue)
   def run_client(output_queue)
     while !stop?
       self.client_socket = new_client_socket
+      metric.increment(:connections)
```
ph: similar comment as https://github.com/logstash-plugins/logstash-input-tcp/pull/56/files#r73949476

jordansissel:
LGTM

acchen97:
For consistency, does it make sense to change `events` to `in`/`out` and add `duration_in_millis` here?

jordansissel:
@acchen97 there is no concept of in/out events on inputs. Inputs only "out" events in this respect. 

For this input, I'm not sure what we'd measure and call `duration_in_millis`.

acchen97:
@jordansissel if that's the case, I think it's fine as is then.

My thought on `duration_in_millis` here would be time spent between the port receiving the event and outputting to the internal queue. Does TCP input simply block additional event entry when the internal queue is full? Any other reasons why the event would be "stuck" in the input? I understand this behavior is different per input, and may be more ubiquitously tracked in like a `time spent waiting on queue` metric along /w persistence later.

PR #60, opened 2016-11-19
return instead of killing entire process

nneul:

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -171,7 +171,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
         # PROXY proto clientip proxyip clientport proxyport
         if pp_info[0] != "PROXY"
           @logger.error("invalid proxy protocol header label", :hdr => pp_hdr)
-          raise IOError
+          return
         else
           proxy_address = pp_info[3]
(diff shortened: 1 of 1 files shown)


jordansissel:
I think the previous code is correct. We should raise an exception, but the behavior should be that the problematic connection is terminated.

By `killing entire process` you mean Logstash dies? Can you show the logs?

nneul:
Nothing traps the IO Error, so the whole listener terminates and that in turn kills logstash itself. It looks like existing code has a catch for ErrNo::ECONNRESET only.

nneul:
would 'raise IOError' and adding a 'rescue' that does nothing - equivalent to handling of EOFError in the current tcp code? Similar for other two plugins (syslog and log4j)?

Is there a better exception to use?

nneul:
(re question on the syslog plugin) The (crash) is completely expected in this case - occurs if you configure the input to "use proxy protocol" and then make a non-proxy-protocol connection to the service. It just isn't exiting properly. It's primarily a question about the proper way to say "this connection is bad, I'm dropping it".  The IOError exception that was being raised resulted in the entire listener/plugin terminating, instead of just the individual connection.

kares:
defunct - exception is expected to be catched at a higher level by LS. simply return-ing does not make sense ...

PR #63, opened 2016-12-02
Make this input plugin capable of doing line delimiting.

jordansissel:
This change keeps the behavior the same, but inverts the previous
implementation which used `fix_streaming_codecs` to automatically
switch from `plain` to `lines` and from `json` to `json_lines`. The
switch was needed because the input wasn't programmed to make
choices about line delimiters -- that job was handled by the `line` and
`json_line` codec.

This particular change was driven by a discussion about improving the
`cef` codec and how I personally didn't want to have both a `cef` and
`cef_lines` codec.

I've had discussions over the past year or two with folks on the
Logstash team where we talked about separating "payload delimiting"
concerns away from the codecs and moving that delimiting functionality
into a new system we tentatively called a "Mill" (credit to @guyboertje
for the term and I think the original idea) after the idea of a sawmill
which cuts raw material (logs) into usable pieces (lumber).

This commit doesn't introduce "mill" as a named concept, but implements
the idea that codecs are probably not the best place for delimiting to
happen. You can feel the weirdness of some codecs doing delimiting in
several places:

* The file input (files are a stream of bytes) emitting *lines* to its codecs.
* The tcp input emitting *bytes* to its codecs
* The fact that `json` and `json_lines` codecs exist and do the same
thing except one does line delimiting also.
* The fact that `plain` and `line` codecs exist and do the same thing
except one does line delimiting also.
* The `fix_streaming_codecs` method that hacks around the previous two
points to force (previously) the `line` codec to be used on the tcp
input if the user mistakenly configured `plain`. Same for JSON.

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -1,6 +1,7 @@
   # encoding: utf-8
 require "logstash/inputs/base"
 require "logstash/util/socket_peer"
+require "logstash/util/buftok"
 
 require "socket"
 require "openssl"
@@ -14,7 +15,7 @@ require "openssl"
 class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   config_name "tcp"
 
-  default :codec, "line"
+  default :codec, "plain"
 
   # When mode is `server`, the address to listen on.
   # When mode is `client`, the address to connect to.
@@ -57,6 +58,9 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   # Useful when the CA chain is not necessary in the system store.
   config :ssl_extra_chain_certs, :validate => :array, :default => []
 
+  # The delimiter for events as bytes are read from the tcp socket.
+  config :delimiter, :validate => :string, :default => "\\n"
+
   HOST_FIELD = "host".freeze
   PORT_FIELD = "port".freeze
   PROXY_HOST_FIELD = "proxy_host".freeze
@@ -65,6 +69,8 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
 
   def initialize(*args)
     super(*args)
+
+    @delimiter = @delimiter.gsub("\\r", "\r").gsub("\\n", "\n")
 
     # monkey patch TCPSocket and SSLSocket to include socket peer
     TCPSocket.module_eval{include ::LogStash::Util::SocketPeer}
@@ -80,7 +86,18 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   end
 
   def register
-    fix_streaming_codecs
+
+    # Delimiters are handled by this input, now, so let's do some backwards
+    # compatibility support by shunting `json_lines` usage to just `json`
+    if @codec.class.config_name == "json_lines"
+      require "logstash/codecs/json"
+      @codec = LogStash::Codecs::JSON.new
+    end
+
+    if @codec.class.config_name == "line"
+      require "logstash/codecs/plain"
+      @codec = LogStash::Codecs::Plain.new
+    end
 
     # note that since we are opening a socket in register, we must also make sure we close it
     # in the close method even if we also close it in the stop method since we could have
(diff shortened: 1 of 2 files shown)


jordansissel:
This change does have a negative impact that codecs providing their own custom delimiting of any kind will probably not work.

I'd like to get closer to having event milling be a configurable concept, but didn't have time tonight for that world shift ;)

breml:
@jordansissel about a year a go I implemented a codec, which is able to chain multiple codecs (something like a meta codec). Maybe this concept would also help for this issue, where a first codec could split the input into lines and the second one cares about the actual format (cef, json, etc.). When using this on an output, they would be chained the other way round, where the event first gets encoded (cef, json, etc.) and gets the proper line ending, if needed.

I had several use cases for this chain codec back then:
* Decouple the output, such that this does not block Logstash, if one output is not able to deliver its events.
* Pre process the event to remove some fields, which should not be sent on one output, but on the others (like black-/whitelisting).
* Some more, I not remember anymore. :thinking:

jordansissel:
@breml When the codecs system was first being designed (Logstash 1.2), I remember having long discussions with @nickethier (and maybe other folks?) about exactly this. We knew chainable codecs were a feature we could do, but ultimately[1], we decided against it.

[1] Assuming my memory is correct, anyway, I remember us deciding not to do this.

We've had internal discussions about composable codecs (chainable, etc), but haven't moved forward on any design.

What I remember (this was like, 3 years ago), was that I was worried that making codecs composable would yield very bad user experience because it would be very very easy to make mistakes (For example, a composition of [line, json] might work, where [json, line] would fail), and I'm generally going to resist features that lay traps for users.

That said, we came up with a compromise, kinda sorta, which allowed composition without increasing risk of failure to users -- a codec's code could use other codecs. This kind of idea lead to `json` + `json_lines` being two different codecs.

-----

I would love to make codec composition available, but I'm not sure how to solve it in a way that doesn't make me worry so much about confusing users.

breml:
@jordansissel Thanks for your detailed answer. Indeed we faced the problem you outline where the ordering of the codecs has to be right, because otherwise it would not work. From my experience we had two different types of codecs (for output plugins):

1. Input: Logstash event, Output: Logstash event (for example: field black-/whitelist, field rename, etc.)
2. Input: Logstash event, Output: "string" (CEF, json, etc.)

May be there would be a third type, which takes a string as input and outputs a string as well (make json_lines from json).

If codecs are used on input plugins, the types would be:

1. Input: "string", Output: string
2. Input: "string", Output: Logstash event
3. Input: Logstash event, Output: Logstash event

If the input and output types would be part of the plugins signature, it would be possible to check for the right chaining of the codecs during the validation of the configuration. I guess this would reduce the impact on the users.

I just openly thinking.

jordansissel:
Indeed. I am also trying to think of ways to solve this. Signatures could help, say, a codec declares it takes bytes and emits also bytes, like the line codec; or a codec that declares it takes bytes and emits events. It's wonky :\

We've also had internal discussions that maybe we don't need codecs anymore -- our idea was to remove `decoding` from codecs and replace them with mills (described above) and move decoding to be a filter instead.

* input delimiting would be handled by a "mill" concept that does things like line, multiline, etc.
* mills would create a simple event where the result is an event having some metadata (who sent it, etc) and the bytes (the line of text)
* decoding part of codecs is basically now just a filter that operates on this 'bytes' field.
* chaining becomes possible because you can run any number of filters in any order already.

You can see how "decoding in a filter" already exists in several places that could arguably be both a codec (today) and a filter: json, multiline, xml, msgpack, csv, etc. JSON filter exists as does JSON codec. Same for multiline.

This kind of transition would let you compose any sequence of "decoding" filters (json, gzip, etc).

jordansissel:
@breml on composing codecs, mills, etc, I'm open to discussion. If you want, open a ticket against elastic/logstash and we can figure things out.

jordansissel:
I'm content to close this, unmerged, and remember it fondly as a small step towards exploring improvements around codecs/filters/mills.

PR #67, opened 2017-04-28
Add log4j2 example to tcp input

jordansissel:
The example aims to help users discover that the tcp input can be used
    to receive data from log4j2 with small configuration changes.

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -11,6 +11,48 @@ require "openssl"
 #
 # Can either accept connections from clients or connect to a server,
 # depending on `mode`.
+#
+# #### Accepting log4j2 logs
+#
+# Log4j2 can send JSON over a socket, and we can use that combined with our tcp
+# input to accept the logs. 
+#
+# First, we need to configure your application to send logs in JSON over a
+# socket. The following log4j2.xml accomplishes this task.
+#
+# Note, you will want to change the `host` and `port` settings in this
+# configuration to match your needs.
+#
+#     <Configuration>
+#       <Appenders>
+#          <Socket name="Socket" host="localhost" port="12345">
+#            <JsonLayout compact="true" eventEol="true" />
+#         </Socket>
+#       </Appenders>
+#       <Loggers>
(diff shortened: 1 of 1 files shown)


jordansissel:
I am not sure if the tcp input docs are the best place for log4j2 examples, but doing this change here was my first thought, so here it is :)

@dedemorton I could use your advice on this -- should this example go anywhere else instead of this plugin docs?

jordansissel:
Testing this with a small test app in Java writing to Logstash:

```
% ~/build/logstash-5.3.0/bin/logstash -e 'input { tcp { port => 12345 codec => json } } filter { date { match => [ "timeMillis", "UNIX_MS" ] } }'

{
  "loggerFqcn"     => "org.apache.logging.log4j.spi.AbstractLogger",
  "level"          => "WARN",
  "endOfBatch"     => false,
  "thread"         => "main",
  "message"        => "OK",
  "threadPriority" => 5,
  "threadId"       => 1,
  "@timestamp"     => 2017-04-27T07:24:03.399Z,
  "port"           => 39010,
  "@version"       => "1",
  "host"           => "127.0.0.1",
  "loggerName"     => "Foo",
  "timeMillis"     => 1493277843399
}
```

jordansissel:
I suppose to answer my own question, if we decide later that this log4j2 example should live somewhere else, we can move it.

suyograo:
this is awesome. LGTM

jordansissel:
Hopefully once published folks will start finding this document as part of googling for logstash + log4j2

dedemorton:
@jordansissel Sounds like you've already resolved this, but for future reference, I think it's OK to have this content here (assuming that the TCP input is the only way to receive log4j2 logs), but if this is a common use case, we might want to make it more visible by talking about it somewhere else. At some point, we should have an overview section (similar to the PR that you reviewed about transforming data) that covers ingesting data. That's where it would make sense to talk about using the TCP input for log4j2 logs...then we could point users to the TCP input plugin docs for the nitty gritty details.

acchen97:
I'm also adding a reference for this in the deploying scaling guide.

PR #71, opened 2017-06-23
#7 Nio Plain Socket Server

original-brownbear:
For #7 

* Nonblocking Java NIO implementation of the event read-loop
* Had to adjust 2 tests slightly to use an actual `write` instead of `puts` on the socket. `puts` triggers two subsequent writes and hence in some cases packets. In combination with non-blocking reads and no backchannel communication, this means that the pattern of `puts`, `flush`, `close` may actually drop some packets if the close comes too quickly
* Put some effort into drying things up between ssl blocking and plain nio reads, but admittedly there is still room for improvement :)


lib/logstash/inputs/tcp.rb
```diff
@@ -154,11 +171,59 @@ def close
...
+    first_read = {}
+    until stop?
+      key = @loop.next_key
+      next unless key
+      if codecs[key].nil?
+        codecs[key] = @codec.clone
+        first_read[key] = true
+      end
+      buffer = key.attachment
+      bytes = Java::byte[buffer.remaining].new
```
original-brownbear: @jakelandis or anyone else taking a look, any input on making these 4 lines more efficient is highly welcome. I refuse to accept that this is the best I can do here to get a Ruby `String` from my direct byte buffer. Then again, the way the `@codec` work, it probably is apart from moving some of the logic below to Java (but that won't get us much either since it's just ran at the beginning of each stream).

original-brownbear:
@jakelandis I don't see what Netty would simplify here? We have such a trivial case of one Socket running accept and the others just reading, I don't think Netty code will be shorter (or more readable) in any meaningful way?

jakelandis:
@original-brownbear 

I am not arguing that in this specific case, Netty is better [1]. I am more concerned about technology convergence and the eventual TLS use case. 

Technology convergence:
Beats input currently uses Netty, and I thought there was desire to consolidate our NIO into something re-usable.  Though in fairness, I believe that discussion started _after_ this was nearly complete.

TLS:
I am speculating that it is considerably more complex to implement TLS directly with socket +nio, then implementing via Netty.  

[1]
_I think there may be an argument to made, that even for this simple case that Netty can provide some value, but would I like to address the first two points above first._

original-brownbear:
@jakelandis 

> I am speculating that it is considerably more complex to implement TLS directly with socket +nio, then implementing via Netty.

It depends imo :) I don't really see the value in Netty end-to-end for these long running connections. Mainly because we have the annoying synchronization over to JRuby that doesn't allow for much buffering in Java anyways (this is where Netty shines and simplifies things in NIO imo).
We def want Netty's warppers for the decoding, but for the end to end selecting and buffering I'd wait until we have a reason to add Netty.
Always note that we do have to do the session handling in Ruby here since the codecs are stateful and Ruby only :( => Netty would also just be firing `select` and `read`.

>  I thought there was desire to consolidate our NIO into something re-usable.

What would that look like though? As I said, we don't have any session handling in Java here, what reusable logic can we provide?

jakelandis:
> It depends imo :) 

I will defer to your opinion since you are much closer to the implementation details. 

>What would that look like though? 

Not sure, I didn't have any concrete proposals, just trying to keep it DRY. Again, I will defer to your opinion If it doesn't makes sense to re-use across the projects.

I checked out your ```nio``` branch and ran it locally, with the following config
```
input { 
tcp {
    port => 55115
	}
}
filter {}
output {  stdout { codec => rubydebug }}
```
and sending the command
```
echo 'hi' | nc localhost 55115
```
I see the message gets accepted via Logstash stdout, but the command line call never returns. (e.g. the command prompt seems to hang).  But when running the same command against master, it works as expected.

original-brownbear:
@jakelandis 

> But when running the same command against master, it works as expected.

Hmm, is this expected? Why should the server go down after the last connection disconnected?

original-brownbear:
@jakelandis  seems like https://github.com/logstash-plugins/logstash-input-tcp/pull/60 is trying to fix the whole process dying from a connection dropping too and rightfully so I guess?

jakelandis:
@original-brownbear  I am not sure I understand your prior comments. 

I ran trace for ```echo 'hi' | nc localhost 55115``` via wire shark and it appears that on [connection termination](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#/media/File:TCP_CLOSE.svg) in the stdio model (master) the client is sending the FIN and the server is responding in kind with the FIN,ACK.  But in the nio (this PR), the client is sending the FIN, but the server does not respond with FIN. 

I believe this results in a half closed connection, which I don't believe is the intent here.

original-brownbear:
@jakelandis hmm yea this is always a fun one in NIO obviously. We could time out connections, but I'm not sure what an appropriate timeout is for our users? Any suggestions? (normally non blocking TCP systems make this configurable, but not sure we want to add that complexity yet)

original-brownbear:
@suyograo @jordansissel any input on what a proper read timeout on client connections would be here from one of you guys maybe? :)

jakelandis:
I think that if the client requests the connection to be closed we should honor it without delay. I am not sure the specifics of how nio would influence that behavior. 

I double checked behavior against the discard and echo server examples from Netty and they seem to honor the full close without an apparent delay. 

I copied the two examples (echo and discard) for easier use: https://github.com/jakelandis/netty-examples

original-brownbear:
@jakelandis 

> I think that if the client requests the connection to be closed we should honor it without delay. I am not sure the specifics of how nio would influence that behavior.

No can do for nonblocking I/O sorry, that's how it works :( 

>  double checked behavior against the discard and echo server examples from Netty and they seem to honor the full close without an apparent delay.

Yea but those are SSL connections?

One plain sockets you can get timeouts from Netty's http://netty.io/4.0/api/io/netty/handler/timeout/ReadTimeoutHandler.html, but there is no portable close event unfortunately. Hence all the pain around `SO_KEEPALIVE` :)

original-brownbear:
maybe we can just un-deprecate this in `tcp.rb`:

```rb

  config :data_timeout, :validate => :number, :default => -1, :deprecated => "This setting is not used by this plugin. It will be removed soon."

```

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Read buffer size.
+     */
+    private static final int BUFFER_SIZE = 8096;
+
+    /**
+     * Queue of {@link SelectionKey} that have data attached and are waiting to be drained by a
+     * JRuby thread.
+     */
+    private final ArrayBlockingQueue<SelectionKey> readable = new ArrayBlockingQueue<>(64);
```
jakelandis: Does this translate in to a max of 64 events in/pending process, else the accepting of new event will block ?

If so, should we bump up the default and/or make this configurable ?

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+public final class InputLoop implements Runnable, Closeable {
+
+    /**
+     * Read buffer size.
+     */
+    private static final int BUFFER_SIZE = 8096;
```
jakelandis: What happens is the payload is larger then 8k ?

original-brownbear: @jakelandis then we only read 8k, no exceptions thrown or anything if that's your worry? :)

jakelandis: so max payload is 8k, else it gets truncated ? 

Seems like a reasonable upperbound for most of our use cases (as I understand them), perhaps we should wrap this in java system.property with 8k as the default for the unexpected use case ?

src/main/java/org/logstash/tcp/InputLoop.java
original-brownbear: @jakelandis yes, but think about this:

Every key that you put in here becomes inactive for receiving data (this could be further improved by using buffer pairs, but I didn't want to go crazy just yet :)). We are only draining this queue using a single thread, 64 already seems like a lot actually.
Guess it depends on your codec and how burst heavy the incoming data is.
To me it seems like this shouldn't make much of a difference with internal socket buffers. We could make it larger though at no cost I think. `1024` would probably look the same and only consume a trivial amount of added memory :)?

src/main/java/org/logstash/tcp/InputLoop.java
original-brownbear: @jakelandis no it doesn't get truncated, it just gets read in two steps (we have additional buffers in Ruby), that's why I'm so careful about what I do to the memory here :D I expect the Ruby side to be the bottleneck here either way :(

src/main/java/org/logstash/tcp/InputLoop.java
original-brownbear: @jakelandis come to think about this, I hate my approach here. Give me a sec to fix :) There is a much more rational way of deciding how long these queues are supposed to be, incoming

original-brownbear:
@jakelandis just as a reminder for me, before we forget :)

* Add timeouts to dropping clients
* Unify buffers to limit overall buffer size
* Block according to buffer size, not `SelectionKey`s

jordansissel:
> any input on what a proper read timeout on client connections 

Read timeouts will cause data loss. I recommend we do not use timeouts.

original-brownbear:
@jordansissel we have need some idle timeouts because we cannot detect if a client closed a session for a non-blocking socket unless we write to it. If we don't time out at all we're leaking file descriptors?

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -0,0 +1,150 @@
...
+        final ByteBuffer buffer = (ByteBuffer) key.attachment();
+        boolean result;
+        try {
+            result = ((ReadableByteChannel) key.channel()).read(buffer) >= 0;
+        } catch (final IOException ignored) {
+            result = false;
+        }
+        buffer.flip();
+        if (buffer.remaining() > 0) {
+            decoder.decode(key);
```
original-brownbear: @jordansissel basically the question is how many workers we want to put on this?

original-brownbear: @jordansissel and in the same go also, how much data do we want to buffer overall before blocking the I/O thread, given that we're CPU bound here for sure.

original-brownbear:
@jakelandis here we go, moved to Netty now that we have the cleaner JRuby handling :)
Left it at the default buffer pool sizing which should be fine imo (64Mb/1k connections as far as I can see) and single threaded processing (seems if we massively buffer here like Netty does, JRuby picks up the pace (fewer individual but larger arrays are set up and collected)).

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -0,0 +1,125 @@
...
+    private final ChannelFuture future;
+
+    /**
+     * Ctor.
+     * @param host Host to bind the listen to
+     * @param port Port to listen on
+     * @param decoder {@link Decoder} provided by Jruby
+     */
+    public InputLoop(final String host, final int port, final Decoder decoder) {
+        worker = new NioEventLoopGroup();
```
jakelandis: This defaults to     
```
DEFAULT_EVENT_LOOP_THREADS = Math.max(1, SystemPropertyUtil.getInt(
                "io.netty.eventLoopThreads", NettyRuntime.availableProcessors() * 2));
```
_which delegates to  Runtime.getRuntime().availableProcessors()_

Is the *2 a concern here ?  I really have no idea what the proper number should be.

original-brownbear: @jordansissel said so right (Jordan please confirm :D)?

I have no idea either, my gut feeling would be `1` to avoid negatively affecting the overall system under pressure, but that's way too slow here :(
Maybe this value is ok, given that the main Queue blocks anyway and we will balance things out this way?

lib/logstash/inputs/decoder_impl.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def decode(channel_addr, data)
+    if @codecs[channel_addr].nil?
+      @codecs[channel_addr] = @codec.clone
+      @first_read[channel_addr] = true
+    end
+    bytes = Java::byte[data.readableBytes].new
+    data.getBytes(0, bytes)
+    tbuf = String.from_java_bytes bytes, "ASCII-8BIT"
```
jakelandis: Is the ASCII-8BIT needed here ? 

I _think_ this will instruct the String object to treat each character as 8bits, which works without question if the original encoding is ASCII7/8 bit. However if the original encoding was something else AND something down the chain tries to do something with the String object (like print it) characters outside the 7/8 bit codec will be (for example printed) incorrect. (another example could be trying split on a unicode character) 

I couldn't find much doc on the exact behavior of JRuby String's implementation, but if it works without the presumption of ASCII-8BIT, I would suggest to remove it.

jakelandis:
@original-brownbear - Changes look really good! 

The two outstanding comments in the review are more questions then requests for change. 

_Also, no more half closed connections :) ...I ran a quick test through wireshark to confirm_

lib/logstash/inputs/decoder_impl.rb
jordansissel: @jakelandis the Codec system in Logstash takes care of byte-to-character handling. The Input doesn't need to handle this. 

In Ruby, there is no "byte array" so folks use a String that has encoding ASCII-8BIT which has some nice properties that string indexing/ranges operate on bytes (octets) instead of characters (1-4 (maybe 6?) bytes in UTF-8)

original-brownbear: @jakelandis `ASCII-8BIT` is raw bytes basically. If it's some other encoding, then that will be enforced down the line by the `@codec`. => `ASCII-8BIT` always works and will never coerce anything about the encoding  :) (Ruby is weird :D I know)

src/main/java/org/logstash/tcp/InputLoop.java
jordansissel: 2*N (2 x cores) seems like a reasonable default.

> be 1 to avoid negatively affecting the overall system under pressure

for most modern systems (in the past decade) you'll have multiple cores available. If you default to 1, the fastest you can process is the top-performance of a single core. With N*2, the fastest you can go is N, and you can't go faster than N. I suspect N*2 is some assumption that some threads may, at times, become blocked by I/O (disk, network, etc) but otherwise not consume CPU.

original-brownbear:
@jakelandis @jordansissel I'm off to bed now, but feel free to merge this later if you're cool with the number of threads (since I know someone is waiting for this fix :)).

lib/logstash/inputs/decoder_impl.rb
```diff
@@ -0,0 +1,63 @@
+# encoding: utf-8
```
jordansissel: We tend to put plugin-specific code in this pattern of file name:

```/lib/logstash/TYPE/NAME/xyz.rb```

For this, I mean: `lib/logstash/inputs/tcp/decoder_impl.rb`

Comment (original-brownbear) [inline lib/logstash/inputs/decoder_impl.rb]: @jordansissel got it let me fix it before bedtime sec :)

File: logstash-input-tcp.gemspec [hunk 8]
```diff
@@ -1,13 +1,14 @@
 Gem::Specification.new do |s|
   s.name          = 'logstash-input-tcp'
-  s.version         = '4.1.2'
+  s.version       = ::File.read('version').split("\n").first
```
jordansissel: Moving the version to a separate file can make it difficult for us to automatically version bump plugins as we do during plugin publishing.

What's the story on moving to a version file? (This can be OK, but I just want to see the story)

spec/inputs/tcp_spec.rb
```diff
@@ -77,11 +77,11 @@
 
     events = input(conf) do |pipeline, queue|
       socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
-      socket.puts("PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r");
+      socket.syswrite("PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r\n")
       socket.flush
       event_count.times do |i|
         # unicode smiley for testing unicode support!
-        socket.puts("#{i} ☹")
+        socket.syswrite("#{i} ☹\n")
```
jordansissel: If you are switching to `syswrite` (instead of `puts`) you do not need to call `flush` (next line down)

spec/inputs/tcp_spec.rb
```diff
@@ -241,11 +241,10 @@
     events = input(conf) do |pipeline, queue|
       event_count.times do |i|
         socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
-        socket.puts("#{i}")
+        socket.syswrite("#{i}\n")
```
jordansissel: No longer need `flush` here.

logstash-input-tcp.gemspec
original-brownbear: @jordansissel I already did it the same way in the user agent input filter , the motivation was just to not duplicate the version between Gemspec and Gradle build file.
Not cool? :)

lib/logstash/inputs/decoder_impl.rb
jordansissel: no rush :)

spec/inputs/tcp_spec.rb
original-brownbear: @jordansissel we do, test fails without it, we don't have a blocking read here and the timing could be unfortunate enough to make this fail without a flush call.
(fails in 2/3 of all runs if I remove the flush here on my Linux box)

lib/logstash/inputs/decoder_impl.rb
original-brownbear: @jordansissel fixed :)

logstash-input-tcp.gemspec
suyograo: I've seen this pattern used in beats, dissect and others as well. https://github.com/logstash-plugins/logstash-input-beats/blob/master/logstash-input-beats.gemspec#L1

Just a side note that we'll need to consolidate these approaches at some point...

spec/inputs/tcp_spec.rb
original-brownbear: @jordansissel I do, see explanation below. This only works by accident (admittedly even `flush` is a bit of an accident but on Loopback it's fine I guess ;D) on a blocking read.

logstash-input-tcp.gemspec
jakelandis: FWIW, Beats input also has a separate VERSION file, and being new, I haven't had an [cognitive friction](https://cogfriction.wordpress.com/about/) during the release process.

jakelandis:
@original-brownbear - don't forget the [release notes](https://github.com/logstash-plugins/logstash-input-tcp/blob/master/CHANGELOG.md)

jordansissel:
> this means that the pattern of puts, flush, close may actually drop some packets if the close comes too quickly

Is this a bug in the test or in the Netty code?

The codec system is designed to accept bytes and emit events. Loss should not be occurring. A single event `Hello world\n` *could* show up as 12 separate TCP PSH packets with 1 byte each, and despite that, the tcp input (today) would correctly wait until the newline before producing an event `{ "message": "Hello world" }`. If this new code does not do this, then there is a bug.

jordansissel:
Can we hold off merging this until the "flush" problem is explained a bit more? Data loss should not be occurring.

original-brownbear:
@jordansissel umm you're right. We have to add those 3 lines for the shutdown flush handling here. Sorry too tired for that today, that will go to tomorrow then ... damn timezones :)

jakelandis:
It looks like socket#puts may be adding in the separator.

https://github.com/jruby/jruby/blob/3766e89176b8e03e407518639c85b66905fee5fa/core/src/main/java/org/jruby/RubyIO.java#L2445

https://github.com/jruby/jruby/blob/3766e89176b8e03e407518639c85b66905fee5fa/core/src/main/java/org/jruby/RubyIO.java#L2519

original-brownbear:
@jakelandis yup it does, but we have another problem here, we're not guaranteed to read till EOF. Haha sorry can't explain it now, almost 1PM :) Will add fix and explaination in ~9h :)

original-brownbear:
@jakelandis @jordansissel ok so this is the deal here :)

1. Closing a socket doesn't guarantee that data will be sent before `close` returns
2. That's why `SO_LINGER` was invented
3. Jruby before 9k does not support setting it https://github.com/jruby/jruby/pull/4119 and throws ```
Java::JavaLang::ClassCastException: org.jruby.RubyClass cannot be cast to org.jruby.ext.socket.Option
```
if I try.
4. Since this also means I can't turn off Nagle and I can actually reproduce like one in 50 runs failing even witht the flush, I'd suggest maybe this:

```
      event_count.times do |i|
        socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
        socket.syswrite("#{i}\n")
        socket.flush
        sleep 0.3
        socket.close
      end
```

0.3s since Nagle is specified at 0.2s :) Or to create a separate issue for this if it's really a problem. This is exclusively an issue with the tests in my opinion.
Further reading: https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable

**Comment by original-brownbear** [comment]
@jordansissel @jakelandis small update on this one, just went over the tests for another project to see how we generally handled things there, seems this is mainly an instability on OSX (though imo, formally (not necessarily in practice), there are no guarantees here unless you flush and linger). Since it's a spec me (and I'm sure many others) would appreciate keeping the flush to have stability for local runs though :)

**Comment by jakelandis** [comment]
@original-brownbear  Trying to understand the need for the ```sleep```, wouldn't the flush ensure that all data is sent prior to the close ? (including the small packets under Nagle's algorithm)

**Comment by original-brownbear** [comment]
@jakelandis 

>  wouldn't the flush ensure that all data is sent prior to the close?

Empirically on OSX the answer is no (at least on my machine :D), but take the suggestion with a grain of salt please (that's why I suggested moving it to a new issue):
1. I did not research what JRuby does here.
2. I can only reproduce this on OSX at a low rate.
3. `0.3` may just be as good as any other value and this has no actual connection to Nagle and is just a result of whatever the `flush` does under the hood in JRuby (+ note there was work between various JRuby versions on exactly the behavior of this ...)

I just think this is a pretty clear cut issue with the spec and not the code, code would imply a Netty bug (imo) right?

**Comment by original-brownbear** [comment]
@jordansissel @jakelandis I take everything back this was a potential issue, the issue was that we weren't properly waiting on Netty shutdown. Fix pushed :)

**Comment by original-brownbear** [comment]
and flushes removed

**Comment by original-brownbear** [comment]
@jordansissel @jakelandis nevermind the above ... no need for the noisy spec changes. Just reverted the spec now that it works either way.

File: src/main/java/org/logstash/tcp/InputLoop.java (hunk 5 above)
Comment (original-brownbear) [inline src/main/java/org/logstash/tcp/InputLoop.java]: @jordansissel @jakelandis on that note, ... we're not done yet I just realized. With this change (having more than one processing thead unlike in the initial solution) we can't use Ruby hashes in in the `DecoderImpl` for concurrent access to the individual session's codecs :( Will fix this soon and ping you guys then.

**Comment by original-brownbear** [comment]
@jakelandis @jordansissel pushed a naive fix for the concurrency issue. I know just brute force using `ConcurrentHashMap` is far from optimal here, but it's not even visible in throughput tests given the other code ... any faster approach would mean changing more code yet again :)

I suggest going with this and adding an issue that asks for a faster approach. (imo just using a session object that contains a reference to that sessions `@codec` will be fine here and require 0 sync)

File: src/main/java/org/logstash/tcp/InputLoop.java [hunk 11]
```diff
@@ -0,0 +1,127 @@
...
+    private final ChannelFuture future;
+
+    /**
+     * Ctor.
+     * @param host Host to bind the listen to
+     * @param port Port to listen on
+     * @param decoder {@link Decoder} provided by Jruby
+     */
+    public InputLoop(final String host, final int port, final Decoder decoder) {
+        worker = new NioEventLoopGroup(1);
```
jakelandis: Did you mean to set this to 1 ?

jakelandis:
There is also this concurrent map that seems to be in some places:
``` 
require 'concurrent'
map = Concurrent::Map.new
```
Though I have no idea in JRuby why to favor that one over just using java util. 

either way +1 to simple fix 

_but please see my comment about the change to number of workers_

original-brownbear:
@jakelandis sorry, can you link the comment again, not sure what you're referring to :(

src/main/java/org/logstash/tcp/InputLoop.java
original-brownbear: @jakelandis ah no, thanks!

original-brownbear:
@jakelandis thread count oversight fixed :)

original-brownbear:
@jakelandis @jordansissel finally done I think :D Moved to a proper `Decoder` per session solution, after all, couldn't reason out why the concurrent map thing failed to stabilize ...  rock solid now in my testing and much easier to understand :)

jakelandis:
@original-brownbear code LGTM, Also, I swear I saw the release notes before ... but not now ?

@jordansissel - Are all of your concerns met ? Also, you may want to validate the most recent concurrent changes to handle the codec since I am not familiar with the underlying codec implementation.

original-brownbear:
@jakelandis thanks! Maybe just F5 the page?, release notes are there I swear :D

original-brownbear:
@jakelandis @jordansissel codec should be much much safer now (also practically is safer now actually), we clone the object for every session now and attach it to the Netty handler. No potentially error prone lookups in some shared cross-session state :)

jordansissel:
I get this warning logged when running:

```
Jun 29, 2017 12:00:58 PM io.netty.util.ResourceLeakDetector reportUntracedLeak
SEVERE: LEAK: ByteBuf.release() was not called before it's garbage-collected. Enable advanced leak reporting to find out where the leak occurred. To enable advanced leak reporting, specify the JVM option '-Dio.netty.leakDetection.level=advanced' or call ResourceLeakDetector.setLevel() See http://netty.io/wiki/reference-counted-objects.html for more information.
```

Tested with this:

clients:

```
% seq 4 | xargs -n1 -P4 sh -c 'seq 10000 | nc localhost 10000' -
```

config:

```
input { tcp { port => 10000 } }
output { null { } }
```

logstash 5.4.0

```
% bin/logstash -f tcp.conf
```

original-brownbear:
@jordansissel wops don't run away fixing in 2 min :D

jordansissel:
TLS seems not to work:

```
[2017-06-29T12:06:44,116][INFO ][logstash.inputs.tcp      ] Starting tcp input listener {:address=>"0.0.0.0:10000"}
[2017-06-29T12:06:44,475][ERROR][logstash.inputs.tcp      ] Could not inititalize SSL context {:exception=>#<OpenSSL::PKey::RSAError: Neither PUB key nor PRIV key:>, :backtrace=>["org/jruby/ext/openssl/PKeyRSA.java:285:in `initialize'", "/home/jls/projects/logstash-input-tcp/lib/logstash/inputs/tcp.rb:295:in `ssl_context'", "/home/jls/projects/logstash-input-tcp/lib/logstash/inputs/tcp.rb:331:in `new_server_socket'", "/home/jls/projects/logstash-input-tcp/lib/logstash/inputs/tcp.rb:140:in `register'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:268:in `register_plugin'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:279:in `register_plugins'", "org/jruby/RubyArray.java:1613:in `each'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:279:in `register_plugins'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:429:in `start_inputs'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:323:in `start_workers'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/pipeline.rb:214:in `run'", "/home/jls/build/logstash-5.4.0/logstash-core/lib/logstash/agent.rb:398:in `start_pipeline'"]}
```

Config:

```
input { tcp {
  port => 10000
  ssl_enable => true
  ssl_cert => "test.crt"
  ssl_key => "test.crt"
} }
output { null { } }
```

Cert/key generation:

```
openssl req -x509  -batch -nodes -newkey rsa:2048 -keyout test.key -subj /CN=localhost -out test.crt
```

original-brownbear:
@jordansissel don't you have to use `test.key` for for the `ssl_key` setting? Buffer leak is fixed now btw.

jordansissel:
Oops, good catch. I was too hasty in my testing of tls.

jordansissel:
tls testing continues ---


```
% openssl s_client -connect localhost:10000
CONNECTED(00000003)
depth=0 CN = localhost
verify error:num=18:self signed certificate
verify return:1
depth=0 CN = localhost
verify return:1
139751240238968:error:140790E5:SSL routines:ssl23_write:ssl handshake failure:s23_lib.c:177:
```

The server seems to be rejecting the connection.

The handshake completes, but then is immediately terminated. Here is what wireshark shows:

image

The TLS handshake seems reasonable. Some odd points:

1) The server is requesting client certificates. It is OK to do this, but given the server was not instructed to verify clients, this could be a signal of the issue
2) The client responds to the CertificateRequest correctly saying "I have no certificates to give you". In this same packet is the Certificate, ClientKeyExchange, ChangeCipherSpec, etc -- so any one of these could be why the server is aborting the connection.

After the last client packet containing the above details, the server immediately terminates the connection (FIN).

jordansissel:
Hmm, maybe strike my previous comment. The default tcp input with logstash 5.4.0 seems to also have this same problem, I need to do more testing to figure out if this PR is causing it.

original-brownbear:
@jordansissel ok, I'm going to bed then :)

jordansissel:
TLS tested working.

Issue #73, opened 2017-06-30
version file incorrectly specified in gemspec definition


robbavey:
As @jakelandis mentions in https://github.com/elastic/logstash/issues/7574#issuecomment-312132206, the version file referenced in https://github.com/logstash-plugins/logstash-input-tcp/blob/master/lib/logstash-input-tcp_jars.rb#L4 is incorrectly specified in the gemspec files list

original-brownbear:
@robbavey yikes :) on it

original-brownbear:
fixed in #74

PR #74, opened 2017-06-30
 #73 fixed broken version file naming in Gemspec

original-brownbear:
for #73 

have to include a lower case version file


suyograo:
LGTM

original-brownbear:
@suyograo thanks, let me wait on Travis will fix re-release then :)

Issue #75, opened 2017-06-30
4.2.1 changes host from IP to hostname


jakelandis:
4.2.1 is the initial release for the Java based NIO implementation. 

It seems that the way Java defines the client address is by the hostname (assuming by using reverse dns).  

The Ruby socket implementation is defined as ```socket.peeraddr[3]``` where ```[3]``` is the [numeric address](https://ruby-doc.org/stdlib-1.9.3/libdoc/socket/rdoc/IPSocket.html). 

The behavior using the Java implementation in 4.2.1:
(message sent from localhost, then from a desktop named link)
```
{
    "@timestamp" => 2017-06-30T17:13:04.737Z,
          "port" => 59539,
      "@version" => "1",
          "host" => "localhost",
       "message" => "hi"
}
{
    "@timestamp" => 2017-06-30T17:18:38.987Z,
          "port" => 56902,
      "@version" => "1",
          "host" => "link",
       "message" => "from link"
}
```
The Ruby 4.1.2 behavior:
```
{
    "@timestamp" => 2017-06-30T17:28:37.359Z,
          "port" => 60348,
      "@version" => "1",
          "host" => "0:0:0:0:0:0:0:1",
       "message" => "hi"
}
{
    "@timestamp" => 2017-06-30T17:28:46.836Z,
          "port" => 56948,
      "@version" => "1",
          "host" => "10.0.0.99",
       "message" => "from link"
}
```

EDIT: fixed the 4.1.2 vs. 4.2.1 swap.  The old way used the IP, the new way is using hostname.

original-brownbear:
@suyograo @jordansissel fix this and new release real quick right?

jordansissel:
++

At the same time, I'd recommend yanking the current release given the regression.

jordansissel:
0) Yank the current release w/ bug
1) PR to fix
2) Release w/ fix

original-brownbear:
@jordansissel on it

original-brownbear:
@jakelandis @jordansissel so we always want reverse lookup, never IP under any circumstances?

original-brownbear:
@jakelandis or is it the other way around and you pasted wrong because currently, it does the reverse lookup? :)

jakelandis:
@original-brownbear - sorry, my dyslexia is kicking in again.. I did paste the examples under the incorrect heading. (will edit).

The passive behavior is always IP.

original-brownbear:
[...]

PR #76, opened 2017-06-30
#75 fixed accidentally reverse-resolving hostnames

original-brownbear:
For #75 
* removed reverse lookup
* added host check to one of the specs
* version 4.2.2 + changelog


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.2.2
```
jakelandis: If we yank 4.2.0 and 4.2.1, then we should probably list those as yanked, and just copy the text from 4.2.0 here.

original-brownbear: @jakelandis all fixed I think :)

jakelandis:
LGTM (with the change log update) and ci passing.

PR #79, opened 2017-07-13
#78 changed behaviour of host field and added new ip_address field

### LINKED ISSUE #78: Optionally perfer hostname of IP for host field
To better consolidate the representation of the ```host``` field,  this plugin should optionally use the hostname in favor of the IPaddress to better match other plugins use of Socket.gethostname. (which is backed by InetAddress.getLocalHost().getHostName() in JRuby). 

I am not sure why TCP input chose IP address. Note the ruby implementation uses [s.peeraddr[3]](https://github.com/logstash-plugins/logstash-input-tcp/blob/v4.1.2/lib/logstash/inputs/tcp.rb#L195) which maps the IP address, and the java version preserves this behavior.  

Perhaps it is latent bug in the TCP input, where the intention was to use hostname, but is an off by one error ?

It seems that other plugins are resulting in the host name not IP, which may cause issues with ES mappings.

original-brownbear:
fixes #78 

* `host` field now does the reverse lookup
* `ip_address` now contains the IP of the peer or is empty for proxy connections (is the behavior for proxy ok here?)
* bumped to `4.3.0` since this is a behavior change and not a patch


original-brownbear:
@suyograo @jakelandis FYI :)

lib/logstash/inputs/tcp.rb
```diff
@@ -107,6 +107,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   config :ssl_extra_chain_certs, :validate => :array, :default => []
 
   HOST_FIELD = "host".freeze
+  HOST_IP_FIELD = "ip_address".freeze
```
suyograo: this needs to be `"[@metdata][ip_address]"` to match beats convention

original-brownbear: @suyograo ah damn one level of nesting slipped by me :D sec fixing :)

original-brownbear:
@suyograo fixed :)

suyograo:
@original-brownbear the version bump needs to be 5.0.0 since it is a breaking change w.r.t. field value + mapping.

LGTM

original-brownbear:
@suyograo right, changing version and merging then, thanks!

jakelandis:
@original-brownbear - we will also to pin LS 5.x to TCP input 4.x

jakelandis:
did a quick test... and all looks great !
_(link is the name of my desktop)_
```
{
    "@timestamp" => 2017-07-13T14:02:41.678Z,
          "port" => 42458,
      "@version" => "1",
          "host" => "link",
      "@metdata" => {
        "ip_address" => "10.0.0.99"
    },
       "message" => "from link"
}
{
    "@timestamp" => 2017-07-13T14:03:13.209Z,
          "port" => 61054,
      "@version" => "1",
          "host" => "localhost",
      "@metdata" => {
        "ip_address" => "0:0:0:0:0:0:0:1"
    },
       "message" => "from localhost"
}
```

original-brownbear:
@jakelandis nice, and thanks for checking!

lib/logstash/inputs/tcp.rb
```diff
@@ -107,6 +107,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   config :ssl_extra_chain_certs, :validate => :array, :default => []
 
   HOST_FIELD = "host".freeze
+  HOST_IP_FIELD = "[@metdata][ip_address]".freeze
```
redbaron4: This should be `"[@metadata]"`. The "a" is missing.

original-brownbear: @redbaron4 oops thanks for spotting this! Will fix this within the next 2 days once I'm back from vacation.

redbaron4: @original-brownbear Looks like this has been fixed in the master (though not yet released)

Issue #59, opened 2016-11-15
TCP : Error closed stream


kaubry:
Hello everybody,

I have been looking for a solution at my problem for days but without any success. 
I tried upgrading to the newly 5.0 stack, setting up 1 VM for every level ( logstash, elastic, kibana), tried to change configuration as well but nothing seems to work.

Basically, I want to syslog drain log from Heroku (https://devcenter.heroku.com/articles/log-drains), send it to my logstash, process it and then send it to my elastic. Everything is working fine for a while until I got this error (which happens once a day) and I need to restart manually in order for logstash to be back running. When this error happens, my CPU and memory go very high (almost 100%) but are low on a normal use.

Here is my error :

```
[...]

PR #84, opened 2017-08-25
Log and continue rather than die when socket accept fails

andrewvc:
Fixes https://github.com/logstash-plugins/logstash-input-tcp/issues/59


jakelandis:
Does the non-SSL implementation behave correctly ?

andrewvc:
@jakelandis I believe it will? See: https://github.com/logstash-plugins/logstash-input-tcp/blob/master/src/main/java/org/logstash/tcp/InputLoop.java#L48

andrewvc:
Refactored the tests in the latest commit to actually check port #s before opening them.

Also, the master / 6.x branches are failing due to https://github.com/elastic/logstash/pull/8097 but should not be blockers.

jakelandis:
Still LGTM

PR #88, opened 2017-09-21
Correcting metadata field name

nradonicich:
Correcting @metdata.ip_address to @metadata.ip_address


karmi:
Hi @nradonicich, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-tcp/pull/88.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

simmel:
This can be closed because #94 resolved it.

Issue #99, opened 2017-12-14
TCP ignores messages without trailing \n v5.6.5 


slatemine:
I suspect this is related to issue #90 and #95 but in moving from logstash 5.4.0 to 5.6.5 we hit this issue, unless a message is newline terminated it seems to get ignored. 

The udp input continued to work as before. 

---

- Version: 5.6.5
- Operating System: ubuntu 16.04 
- Config File (if you have sensitive info, please remove it):
```
input {
  tcp {
    port => "514"
    type => "syslog"
  }
  udp {
    port => "514"
    type => "syslog"
  }
}
```
- Steps to Reproduce:
Toggle comment on line `#msg = "%s\n"%(msg)`
```
#!/usr/bin/env python
import json, datetime
import socket
import logging
from collections import OrderedDict
logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger(__name__)

ts = datetime.datetime.now()
time = ts.isoformat()
obj = {"another":"panda", "field": time}
evt = json.dumps(obj)
logger.info(evt)

dic = OrderedDict()
dic['Event'] = obj

msg = json.dumps(dic, sort_keys=False, ensure_ascii=True)
#msg = "%s\n"%(msg)
msg = msg.encode()
logger.info((msg,time))

socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
try:
    socket.connect(('localhost',514))
    socket.send(msg)
finally:
    socket.close()
```

slatemine:
Codec seems to be lines, even when I try to force it to plain, I wonder it that is related.  
```
[DEBUG][logstash.codecs.line     ] config LogStash::Codecs::Line/@charset = "UTF-8"
[DEBUG][logstash.codecs.line     ] config LogStash::Codecs::Line/@id = "line_b369f23e-4ff4-4656-a3a7-0515fb865645"
[DEBUG][logstash.codecs.line     ] config LogStash::Codecs::Line/@enable_metric = true
[DEBUG][logstash.codecs.line     ] config LogStash::Codecs::Line/@delimiter = "\n"
```

danhermann:
This issue should be resolved as of the commit referenced above and released in v5.0.3 of the TCP input plugin.

slatemine:
v5.0.3 confirmed to be working correctly in ES 5.6.5, but I had to update Gemfile to allow the input to advance passed v4xx.

Issue #95, opened 2017-12-12
TCP input is silently broken in Logstash 5.6.x


pietervogelaar:
We have a Sensu servers that sends Sensu events with a TCP handler to Logstash. This worked fine until Logstash version 5.5.3, but it's broken in 5.6.3 and also 5.6.5. So probably the whole 5.6.x range. However no single error is logged and if I use a Logstash file output, no data is written to it with Logstash 5.6.x.

So it seems to be silently broken.

Sensu server (/etc/sensu/conf.d/handlers/logging_acc.json):
```
{
  "handlers": {
    "logging_acc": {
      "command": null,
      "type": "tcp",
      "socket": {
        "host": "logstash-host.example.org",
        "port": 5555
      },
      "filters": [
        "state_change",
        "non_production_environment"
      ],
      "severities": [
        "ok",
        "warning",
        "critical",
        "unknown"
      ],
      "handle_flapping": false,
      "handle_silenced": false
    }
  }
}
```

Logstash server (/etc/logstash/conf.d/10_input_sensu)
```
input {
  tcp {
    port => 5555
    type => "sensu"
  }
}
```

danhermann:
This issue should be resolved as of the commit referenced above and released in v5.3.0 of the TCP input plugin.

PR #101, opened 2017-12-18
Flush codec after TCP client disconnects.

### LINKED ISSUE #90: regression: last message via TCP input gets lost using multiline codec
when using TCP input together with multiline codec, the message sent to logstash prior to ending the TCP connection is lost

**reproducing**

Dockerfile
```
# works
#FROM docker.elastic.co/logstash/logstash:5.5.3

# broken
FROM docker.elastic.co/logstash/logstash:5.6.0

RUN echo xpack.monitoring.enabled: false >> /usr/share/logstash/config/logstash.yml

COPY pipeline /usr/share/logstash/pipeline
```

logstash configuration (in file pipeline/logstash.conf):
```
input {
    tcp {
        port => 8888
        codec => multiline {
            pattern => "^\s"
            what => "previous"
        }
    }
}

output {
    stdout {
        codec => rubydebug {}
    }
}
```

running logstash in container
```
docker build -t logstash_test .
docker run -ti -p 9600:9600 -p 8888:8888 logstash_test
```

sending data
```
echo -e a\\n 1\\n 2\\nb\\n 1 | nc localhost 8888
```

running logstash 5.5.3, both messages arrive.
with logstash 5.6.0 the last message gets lost

**versions used**

- docker.elastic.co/logstash/logstash:5.5.3
logstash-input-tcp (4.1.2)
logstash-codec-multiline (3.0.7)

- docker.elastic.co/logstash/logstash:5.6.0
logstash-input-tcp (4.2.3)
logstash-codec-multiline (3.0.7)

danhermann:
Relates to #95 and #99

Fixes #90

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -188,6 +188,12 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
     end
   end
 
+  def flush_codec(codec, client_ip_address, client_address, client_port)
+    codec.flush do |event|
+      enqueue_decorated(event, client_ip_address, client_address, client_port)
+    end
+  end
+
   private
 
   RUN_LOOP_ERROR_MESSAGE="TCP input server encountered error"
@@ -278,10 +284,7 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   ensure
     # catch all rescue nil on close to discard any close errors or invalid socket
     socket.close rescue nil
-
-    codec.respond_to?(:flush) && codec.flush do |event|
-      enqueue_decorated(event, client_ip_address, client_address, client_port)
-    end
+    flush_codec(codec, client_ip_address, client_address, client_port)
   end
 
   def enqueue_decorated(event, client_ip_address, client_address, client_port)
--- a/lib/logstash/inputs/tcp/decoder_impl.rb
+++ b/lib/logstash/inputs/tcp/decoder_impl.rb
@@ -25,8 +25,12 @@ class DecoderImpl
                        @proxy_address, @proxy_port, tbuf)
   end
 
-  def copy()
+  def copy
     DecoderImpl.new(@codec.clone, @tcp)
+  end
+
+  def flush
+    @tcp.flush_codec(@codec, @ip_address, @address, @port)
   end
 
   private
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -269,6 +269,41 @@ describe LogStash::Inputs::Tcp do
     end
   end
 
+  it "should flush codec after client disconnects" do
+    # verifies fix for https://github.com/logstash-plugins/logstash-input-tcp/issues/90
+    conf = <<-CONFIG
+      input {
+        tcp {
+          port => #{port}
+          codec => multiline {
+              pattern => "^\s"
+              what => "previous"
+          }
+        }
+      }
+    CONFIG
+
+    data = "a\n 1\n 2\nb\n 1"
+    event_count = 2
+
+    events = input(conf) do |pipeline, queue|
+      socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
+      socket.puts(data)
+      socket.close
+
+      # If the codec is not properly flushed, there will be only one event and the second call to queue.pop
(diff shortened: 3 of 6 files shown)


jordansissel:
@danhermann https://travis-ci.org/logstash-plugins/logstash-input-tcp/jobs/318335342 -- 3 tests failed.

danhermann:
@jordansissel, the tests all pass now that the timing issues are resolved.

lib/logstash/inputs/tcp.rb
```diff
@@ -188,6 +188,12 @@ def decode_buffer(client_ip_address, client_address, client_port, codec, proxy_a
     end
   end
 
+  def flush_codec(codec, client_ip_address, client_address, client_port)
+    codec.respond_to?(:flush) && codec.flush do |event|
```
colinsurprenant: this is not necessary, all codecs respond to `flush`, per https://github.com/elastic/logstash/blob/a3f6da3c3a61470ce510be0ea9ef2b6c07d47d5a/logstash-core/lib/logstash/codecs/base.rb#L73-L77  since a long time.

spec/inputs/tcp_spec.rb
```diff
@@ -269,6 +269,39 @@ def get_port
...
+
+      event_count.times.collect {
+        # use non-blocking pop with sleep so that a missing queue item doesn't cause the test to hang
+        # if a "queue empty" error occurs here, the codec was not properly flushed
+        sleep(2)
+        queue.pop(true)
+      }
+    end
+
+    insist { events.length } == event_count
```
colinsurprenant: I was decided to drop the usage of `insist` for all new specs. We should use the rspec `expect` syntax .

danhermann:
@colinsurprenant, let me know if these changes address your comments.

lib/logstash/inputs/tcp/decoder_impl.rb
```diff
@@ -29,6 +29,10 @@ def copy()
     DecoderImpl.new(@codec.clone, @tcp)
   end
 
+  def flush()
```
colinsurprenant: nit: we ususally don't put `()` on no-arg methods ...

spec/inputs/tcp_spec.rb
```diff
@@ -269,6 +269,39 @@ def get_port
...
+
+    data = "a\n 1\n 2\nb\n 1"
+    event_count = 2
+
+    events = input(conf) do |pipeline, queue|
+      socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
+      socket.puts(data)
+      socket.close
+
+      event_count.times.collect {
```
colinsurprenant: nit: multi-lines blocks should use `do` ... `end`. we try to use `{` ...  `}` for single-line blocks.

spec/inputs/tcp_spec.rb
```diff
@@ -269,6 +269,39 @@ def get_port
...
+        # use non-blocking pop with sleep so that a missing queue item doesn't cause the test to hang
+        # if a "queue empty" error occurs here, the codec was not properly flushed
+        sleep(2)
+        queue.pop(true)
+      }
+    end
```
colinsurprenant: I am not sure I like the hard sleep here, since it will basically force that test to take 4 seconds and the 2 seconds is arbitrary here, could be too much but could be not enough depending, we don't know.
since we are only testing on the number of events maybe using a spin loop with a timeout like
```
Timeout.timeout(timeout) do
  sleep(0.1) until queue.size >= 2
end
```
WDYT?

danhermann: That sounds like it would work better. I wrapped the blocking calls to `queue.pop` with a timeout which should accomplish the same thing.

spec/inputs/tcp_spec.rb
```diff
@@ -269,6 +269,41 @@ def get_port
...
+      socket.close
+
+      # If the codec is not properly flushed, there will be only one event and the second call to queue.pop
+      # will block indefinitely. Wrapping this with a timeout ensures that failure mode does not hang the
+      # test.
+      Timeout.timeout(5) do
+        event_count.times.collect do
+          queue.pop
+        end
+      end
```
colinsurprenant: 👍

colinsurprenant:
LGTM! (did not test). Good job, thanks @danhermann!

PR #102, opened 2017-12-19
Eliminates one class of frequently-encountered timing issues with tests

danhermann:


jordansissel:
LGTM

Thank you for the explanatory comment on each .sort_by. This will help our future selves understand the decision! 👍

PR #106, opened 2018-01-09
Port server-mode SSL to Netty.

### LINKED ISSUE #104: Finish migration to Netty (TLS still uses Ruby Sockets)
https://github.com/logstash-plugins/logstash-input-tcp/blob/master/lib/logstash/inputs/tcp.rb#L140

Let's finish the migration to use Netty for this plugin for both plaintext and tls modes.

danhermann:
The `SslOptions.toSslContext()` method accommodates all the SSL options from the Ruby implementation. Some simple perf tests with the [tcpkali](https://github.com/satori-com/tcpkali) tool showed that the Netty SSL implementation ranged from roughly equal to ~30% faster than the Ruby SSL implementation depending on the test setup. In general, the performance of the Netty implementation exceeded the Ruby implementation when the number of TCP clients exceeded ~500.

After talking with @jordansissel and @jsvd, I removed two integration tests for the previous Ruby SSL implementation for the following reasons:

* The "should log the error on accept" test was removed because the logger that now logs this error is in Java and no longer easily accessible via rspec. The presence of this log message was verified via manual testing.

* The "connection was healthy but now has garbage or corruption" test was removed partly because its success criteria was based on the threading model of the Ruby SSL implementation and partly because it's testing core SSL behavior which should be covered by Netty itself.

Fixes #104.

### DIFF
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -402,29 +402,6 @@ describe LogStash::Inputs::Tcp do
                 raise StandardError, "blah"
               end
             end
-
-            it "should log the error on accept" do
-              allow(input.logger).to receive(:error).with(any_args)
-              
-              stop = Thread.new {
-                sleep 2
-                input.do_stop
-              }
-              expect do
-                input.run(Queue.new)
-              end.not_to raise_error
-
-              expect(input.logger).to have_received(:error).with(
-                ::LogStash::Inputs::Tcp::RUN_LOOP_ERROR_MESSAGE,
-                :message => "blah",
-                :class => "StandardError",
-                :backtrace => anything
-              ).at_least(:once)
-
-              stop.join
-              # Wait for stop to actually happen
-              sleep 1
-            end
           end
 
           context "that sends garbage instead of TLS handshake" do
@@ -440,37 +417,6 @@ describe LogStash::Inputs::Tcp do
               client.flush
               Thread.new { sleep(1); client.close }
             end
-            it "should not negatively impact the plugin" do
-              # TODO(sissel): Look for a better way to detect this failure besides a sleep/wait.
-              result = input_task.thread.join(0.5)
-              expect(result).to be_nil
-            end
-          end
-
-          context "connection was healthy but now has garbage or corruption" do
-            let!(:input_task) { Stud::Task.new { input.run(queue) } }
-            let(:tcp) { TCPSocket.new("127.0.0.1", port) }
-            let(:sslcontext) { OpenSSL::SSL::SSLContext.new }
-            let(:sslsocket) { OpenSSL::SSL::SSLSocket.new(tcp, sslcontext) }
-            let(:max_length) { 1000 }
-            let(:garbage) { Flores::Random.iterations(max_length).collect { Flores::Random.integer(1...255) }.pack("C*") }
-
-            before do
-              sslcontext.cert = certificate
-              sslcontext.key = key
-              sslcontext.verify_mode = OpenSSL::SSL::VERIFY_NONE
-
-              sslsocket.connect
-              sslsocket.write("Hello world\n")
-
-              # Assertion to verify this test is actually sending something.
-              expect(garbage.length).to be > 0
-              tcp.write(garbage)
-              tcp.flush
-              sslsocket.close
-              tcp.close
-            end
-
             it "should not negatively impact the plugin" do
               # TODO(sissel): Look for a better way to detect this failure besides a sleep/wait.
               result = input_task.thread.join(0.5)
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,7 @@ dependencies {
   compile 'io.netty:netty-all:4.1.18.Final'
   compile 'io.netty:netty-tcnative-boringssl-static:2.0.7.Final'
   compile group: 'commons-io', name: 'commons-io', version: '2.5'
+  compile 'org.apache.logging.log4j:log4j-core:2.9.1'
 }
 
 task vendor(dependsOn: shadowJar) << {
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -11,6 +11,7 @@ require "socket"
 require "openssl"
 
 java_import org.logstash.tcp.InputLoop
+java_import org.logstash.tcp.SslOptions
 
 # Read events over a TCP socket.
 #
@@ -137,22 +138,24 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
 
     @logger.info("Starting tcp input listener", :address => "#{@host}:#{@port}", :ssl_enable => "#{@ssl_enable}")
     if server?
-      if @ssl_enable
-        self.server_socket = new_server_socket
-      else
-        @loop = InputLoop.new(@host, @port, DecoderImpl.new(@codec, self))
-      end
+      ssl_options = SslOptions.builder
+        .set_is_ssl_enabled(@ssl_enable)
+        .set_should_verify(@ssl_verify)
+        .set_ssl_cert(@ssl_cert)
+        .set_ssl_key(@ssl_key)
+        .set_ssl_key_passphrase(@ssl_key_passphrase.value)
+        .set_ssl_extra_chain_certs(@ssl_extra_chain_certs.to_java(:string))
+        .build
+
+      @loop = InputLoop.new(@host, @port, DecoderImpl.new(@codec, self), ssl_options,
+                            @logger.to_java(org.apache.logging.log4j.Logger))
     end
   end
 
   def run(output_queue)
     @output_queue = output_queue
     if server?
-      if @ssl_enable
-        run_ssl_server
-      else
-        @loop.run
-      end
+      @loop.run
     else
       run_client()
     end
@@ -196,30 +199,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
 
   private
 
-  RUN_LOOP_ERROR_MESSAGE="TCP input server encountered error"
-  def run_ssl_server()
-    while !stop?
-      begin
-        socket = add_connection_socket(server_socket.accept)
-        # start a new thread for each connection.
-        server_connection_thread(socket)
-      rescue OpenSSL::SSL::SSLError => e
-        # log error, close socket, accept next connection
-        @logger.debug? && @logger.debug("SSL Error", :exception => e, :backtrace => e.backtrace)
-      rescue => e
-        @logger.error(
-          RUN_LOOP_ERROR_MESSAGE, 
-          :message => e.message,
-          :class => e.class.name,
-          :backtrace => e.backtrace
-        )
-      end
-    end
-  ensure
-    # catch all rescue nil on close to discard any close errors or invalid socket
-    server_socket.close rescue nil
-  end
-
   def run_client()
     while !stop?
       self.client_socket = new_client_socket
@@ -228,17 +207,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   ensure
     # catch all rescue nil on close to discard any close errors or invalid socket
     client_socket.close rescue nil
-  end
-
-  def server_connection_thread(socket)
-    Thread.new(socket) do |s|
-      begin
-        @logger.debug? && @logger.debug("Accepted connection", :client => s.peer, :server => "#{@host}:#{@port}")
-        handle_socket(s)
-      ensure
-        delete_connection_socket(s)
-      end
-    end
   end
 
   def handle_socket(socket)
@@ -325,17 +293,6 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
       cert_store.add_file(cert)
     end
     cert_store
-  end
-
-  def new_server_socket
-    begin
-      socket = TCPServer.new(@host, @port)
-    rescue Errno::EADDRINUSE
-      @logger.error("Could not start TCP server: Address in use", :host => @host, :port => @port)
-      raise
-    end
-
-    @ssl_enable ? OpenSSL::SSL::SSLServer.new(socket, ssl_context) : socket
   end
 
   def new_client_socket
--- a/src/main/java/org/logstash/tcp/InputLoop.java
+++ b/src/main/java/org/logstash/tcp/InputLoop.java
@@ -11,8 +11,10 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.handler.ssl.SslContext;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
+import org.apache.logging.log4j.Logger;
 
 import java.io.Closeable;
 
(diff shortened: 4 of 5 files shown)


spec/inputs/tcp_spec.rb
```diff
@@ -402,29 +402,6 @@ def get_port
                 raise StandardError, "blah"
               end
             end
-
-            it "should log the error on accept" do
-              allow(input.logger).to receive(:error).with(any_args)
-              
-              stop = Thread.new {
```
colinsurprenant: no sure I understand why we need to spawn a thread here?

spec/inputs/tcp_spec.rb
```diff
@@ -402,29 +402,6 @@ def get_port
...
-              expect(input.logger).to have_received(:error).with(
-                ::LogStash::Inputs::Tcp::RUN_LOOP_ERROR_MESSAGE,
-                :message => "blah",
-                :class => "StandardError",
-                :backtrace => anything
-              ).at_least(:once)
-
-              stop.join
-              # Wait for stop to actually happen
-              sleep 1
```
colinsurprenant: instead of sleeping here, couldn't we just do a `input_task.wait` ?

colinsurprenant:
@danhermann thanks for this, changes look good, left a few  minor comments on the specs. Other than that it is really hard for me to appreciate the SSL-related changes as I have zero experience with that + Netty. Once the minor comments are addressed I will be ok to LGTM but maybe @jordansissel might want to take a look since he has experience with SSL support?

spec/inputs/tcp_spec.rb
danhermann: @colinsurprenant, that is one of the two integration tests that I deleted as part of this PR. I added some extra detail to the PR description to explain why those two tests were removed.

colinsurprenant: uh ... wtf...  :( of course.

jordansissel:
This is on my list for tomorrow to test.

jordansissel:
Testing:

* `bin/logstash -e 'input { tcp { port => 5000 ssl_enable => true ssl_key => "/tmp/lumberjack.key" ssl_cert => "/tmp/lumberjack.crt"  } } '`
* `openssl s_client -connect localhost:5000` 

Result:

* this pr: data accepted
* current release: tls handshake rejected because no certificate was provided

The default value of `ssl_verify` is `true`, so the current release is behaving correctly by rejecting my connection for being untrusted. This occurs both when the client provides no certificate and when the client provides an untrusted certificate.

I believe this is a bug in the PR.

jordansissel:
Here's the log from tealess (a TLS tool I am working on):

Current version of Logstash which rejects by default (ssl_verify => true):
```
% cli/build/install/cli/bin/cli connect --capath /tmp/lumberjack.crt localhost:5001
localhost resolved to 2 addresses

All SSL/TLS connections failed.

Failure: class javax.net.ssl.SSLHandshakeException - The server rejected our handshake because it did not trust our certificate.
  localhost/127.0.0.1:5001
  localhost/0:0:0:0:0:0:0:1:5001
Here is a network log prior to the handshake failure:
  [Output] ClientHello[version: TLS 1.2, 36 cipher suites; suites: <omitted>]
  [Input] ServerHello[TLS 1.2, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, keyex ECDHE_RSA]
  [Input] CertificateMessage[1 certificates]
  [Input] ServerKeyExchange
  [Input] CertificateRequestMessage(certificate types: [RSASign, DSSSign, ECDSASign])
  [Input] ServerHelloDone
  [Output] CertificateMessage[0 certificates]
  [Output] ClientKeyExchange
  [Input] Alert[Fatal, BadCertificate]
```

And when providing an untrusted certificate:

```
Failure: Network interrupted while handshaking: class java.net.SocketException: Broken pipe (Write failed)
The server may have rejected our handshake because it did not trust our certificate?
Here is a network log prior to the handshake failure:
  [Output] ClientHello[version: TLS 1.2, 36 cipher suites; suites: <omitted>]
  [Input] ServerHello[TLS 1.2, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, keyex ECDHE_RSA]
  [Input] CertificateMessage[1 certificates]
  [Input] ServerKeyExchange
  [Input] CertificateRequestMessage(certificate types: [RSASign, DSSSign, ECDSASign])
  [Input] ServerHelloDone
  [Output] CertificateMessage[1 certificates]
  [Output] ClientKeyExchange
```

jordansissel:
Come to think of it, I don't see any option in the TCP input for specifying the set of trusted authorities (no "certificate authorities" option or similar)...

I have no idea what `ssl_verify => true` is supposed to verify against. This is a very strange plugin.

jakelandis:
I assume it validates against the default java trust store (which on some platforms defaults to the OS trust store), and if not there then check against [ssl_extra_chain_certs](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-tcp.html#plugins-inputs-tcp-ssl_extra_chain_certs). (note- i have not read the source)

There is an theme across our TLS settings ... none of them match :(  this seems the oddest duck however.

danhermann:
@jakelandis, that was my (possibly mistaken) understanding as well. I have not yet tried your suggestion of  a custom trust manager. I plan to do that early next week after I get back.

jordansissel:
```
% bin/logstash -e 'input { tcp { port => 5000 ssl_enable => true ssl_key => "/tmp/lumberjack.key" ssl_cert => "/tmp/lumberjack.crt" ssl_verify => true ssl_extra_chain_certs => "/tmp/lumberjack.crt" } } '
```

Testing this PR:
* ❌ with no certificate: client is accepted
* ✅ with /tmp/lumberjack.crt as client certificate: client is accepted
* ❌  with a differeng (untrusted) client certificate: client is accepted.

Testing this Logstash 6.2.0 build candidate:
* ✅ with no certificate: rejected
* ✅ with /tmp/lumberjack.crt as client certificate: client is accepted
* ✅ with a differeng (untrusted) client certificate: rejected

I think this PR breaks some things about the trust.

When `ssl_verify` is true, the tcp input must reject the following:
1) The client provides no certificate
2) The client provides an untrusted certificate

danhermann:
@jordansissel, I misunderstood what `ssl_verify` was supposed to do. My latest commit adds the behavior you were expecting in your test.

One difference in behavior that I noticed while testing locally is that if missing or invalid certs or keys are supplied, the Ruby TLS implementation fails during `register` and prevents Logstash startup because it attempts to construct the relevant TLS classes then. The Netty implementation doesn't construct those classes until a client attempts to connect and therefore does not prevent Logstash from starting up. If the former behavior is preferred, some additional work will be required to validate the SSL options on startup rather than at client connection time.

jordansissel:
> The Netty implementation doesn't construct those classes until a client attempts to connect 

It may be enough to build the SSLContext which should be enough work to provide validation (valid SSL settings, certificate/trust loading, etc)

danhermann:
I cannot reproduce the 5.6 test failure on this PR, but the diagnostic logging shows that Netty considers the supplied key file to be invalid in which case LS should explicitly fail to start:

`File does not contain valid private key: /tmp/studtmp-f874774e23526348e622cd4b563cf9cf9a7be66879dbfbe6491381ed9b47`

jordansissel:
Chatted with @danhermann on zoom about the [travis failure](https://travis-ci.org/logstash-plugins/logstash-input-tcp/jobs/338126591).

We can only reproduce this failure on Travis. Neither Dan nor myself can reproduce it, even with the same random seed, logstash version, etc.

* `bundle exec rspec` w/ Logstash 5.6.7 passes tests for me.
* Building the gem and installing in Logstash 5.6.7 successfully starts w/ ssl enabled (the travis test failure would indicate this should not succeed)

Therefore, we're gonna ignore the travis failure.

----

In testing the latest commit on this pr (29986f9), the same ssl trusting-everything behavior seems present, and @danhermann is working on it.

jordansissel:
From my previous comment, I had stale java build results so i was using old code. Oops.

I can reproduce the travis failure:

The file itself *is* a valid RSA key. I tested by using `openssl rsa -in /some/path` to read it and confirm.

I'm not sure why or how it is failing.

jordansissel:
Some results and tests that lead us to still remain confused --


```
Certificate: /tmp/studtmp-71b25760ee45c28af1567d289d4701b974e46892934dc56cbe6c1b3fe170
Key: /tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089
Passphrase: null
Passphrase? true
java.lang.IllegalArgumentException: File does not contain valid private key: /tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089
````

Ok so we believe we are invokding SslContextBuilder.forServer() correctly, yet, it crashes telling us the key is not valid. Hypothesis (testable!) is that we are passing somehow an non-null value for the ssl passphrase (3rd arg of SslContextBuilder) but we confirm it is null in the above output!!

Testing a similar path in `bin/logstash --interactive pry`:

```
[1] pry(#<LogStash::Runner>)> z = LogStash::Plugin.lookup("input", "tcp").new("ssl_enable" => true, "ssl_cert" => "/tmp/studtmp-71b25760ee45c28af1567d289d4701b974e46892934dc56cbe6c1b3fe170", "ssl_key" => "/tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089", "port" => 5001, "ssl_key_passphrase" => LogStash::Util::Password.new(nil)).register
=> #<Java::OrgLogstashTcp::InputLoop:0x7a6322e2>
[2] pry(#<LogStash::Runner>)> _.run
[2018-02-07T01:30:22,882][ERROR][logstash.inputs.tcp      ] Error in Netty input handler: java.lang.IllegalArgumentException: File does not contain valid private key: /tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089

^^ Fails, but should work.

[1] pry(#<LogStash::Runner>)> require "logstash/inputs/tcp"
=> true
[7] pry(#<LogStash::Runner>)> java_import "io.netty.handler.ssl.SslContextBuilder"
=> [Java::IoNettyHandlerSsl::SslContextBuilder]
[11] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new("/tmp/studtmp-71b25760ee45c28af1567d289d4701b974e46892934dc56cbe6c1b3fe170").to_java, java.io.File.new("/tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089"), nil)
=> #<Java::IoNettyHandlerSsl::SslContextBuilder:0x6c36ee80>

^^ successfully loads the certificate and key and appears to be, based on the code, what is being invoked. Yet, it fails.
```

In debugging, we confirmed that the getSslKeyPassphrase() returns `null` in the Java side. I have no idea why this is failing at this time.

We're going to sleep on this and hopefully an epiphany waits us in the morning.

**Comment by jordansissel** [comment]
I wonder if this is just bad interaction between Netty and Flores::PKI's generated private keys? (Test suite problem?)

```

# Trying Netty's SslContextBuilder on a cert+key generated by Flores::PKI
# This fails:
[18] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
Java::JavaLang::IllegalArgumentException: File does not contain valid private key: /tmp/studtmp-c9c94ae999e9a11669ab1778ef4d5714c9c133399323a2dc0770db85f089
from io.netty.handler.ssl.SslContextBuilder.keyManager(io/netty/handler/ssl/SslContextBuilder.java:251)

# Loading the private key with JRuby's OpenSSL::PKey::RSA works
[20] pry(#<LogStash::Runner>)> OpenSSL::PKey::RSA.new(File.read(key))
=> -----BEGIN RSA PRIVATE KEY-----
...
```


```
% openssl genrsa -out test.key; openssl req -x509  -batch -key test.key -out test.crt -subj /CN=localhost

[24] pry(#<LogStash::Runner>)> puts key, cert
/tmp/test.key
/tmp/test.crt

[25] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
Java::JavaLang::IllegalArgumentException: File does not contain valid private key: /tmp/test.key
from io.netty.handler.ssl.SslContextBuilder.keyManager(io/netty/handler/ssl/SslContextBuilder.java:251)
```

Maybe Netty just dislikes this key format? I think the format Flores::PKI is generating is PKCS#1?

```
# Convert `test.key` to pkcs8 and save as `test.pkcs8.key`
% openssl pkcs8 -topk8 -in test.key -nocrypt > test.pkcs8.key


# pkcs8 format private key works:
[26] pry(#<LogStash::Runner>)> key = "/tmp/test.pkcs8.key"; cert = "/tmp/test.crt"
=> "/tmp/test.crt"
[27] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
=> #<Java::IoNettyHandlerSsl::SslContextBuilder:0x40d482c8>

# pkcs1 format private key fails:
[28] pry(#<LogStash::Runner>)> key = "/tmp/test.key"; cert = "/tmp/test.crt"
=> "/tmp/test.crt"
[29] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
Java::JavaLang::IllegalArgumentException: File does not contain valid private key: /tmp/test.key
from io.netty.handler.ssl.SslContextBuilder.keyManager(io/netty/handler/ssl/SslContextBuilder.java:251)
```

**Comment by jakelandis** [comment]
@jordansissel - could it be related to https://github.com/jordansissel/ruby-flores/issues/9 and https://github.com/jordansissel/ruby-flores/pull/10 ?

I had to monkey patch the tests for beats input a while back.

**Comment by jordansissel** [comment]
@jakelandis Good idea, though I think that's unrelated. Those two issues are about certificates, and this failure is about private key file formats. TLS is a scream 😱  

I think Ruby OpenSSL has no api for creating PKCS#8 data, so maybe we are SOL, here, for now. Maybe the right solution is to do this --

1) Stop using Flores::PKI for now. Instead, use `openssl` to create a key+cert we can commit into the specs as a fixture.
2) Outside the scope of this PR, find a way to make Flores::PKI or something like it work again with something like Netty which needs PKCS#8 private keys.

**Comment by jordansissel** [comment]
Based on research so far:

1) This failure is reproducible
2) It seems to be primarily becuase `Flores::PKI.generate` produces a private key in format PKCS#1.
3) Netty wants private keys in format PKCS#8

Proposal: Create a key+cert pair (in the correct format) and commit to this repo as part of this PR and use this key material in the tests instead of using Flores::PKI

**Comment by jordansissel** [comment]
Tested the following on Logstash 5.6.7 and 6.2.0:

```
[13] pry(#<LogStash::Runner>)> key = "/tmp/test.pkcs8.key"; cert = "/tmp/test.crt"
=> "/tmp/test.crt"
[14] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
=> #<Java::IoNettyHandlerSsl::SslContextBuilder:0x455cdfab>
[15] pry(#<LogStash::Runner>)> key = "/tmp/test.key"; cert = "/tmp/test.crt"
=> "/tmp/test.crt"
[16] pry(#<LogStash::Runner>)> SslContextBuilder.forServer(java.io.File.new(cert), java.io.File.new(key))
=> #<Java::IoNettyHandlerSsl::SslContextBuilder:0x2d29ab00>
```

The PKCS1 ("/tmp/test.key") fails on 5.6.7. Succeeds on 6.2.0!

Main difference is:
* 5.6.7: netty 4.1.18
* 6.2.0: netty 4.1.3

Maybe?

**Comment by danhermann** [comment]
I rebased this PR after the Netty version was bumped to 4.1.18 and the test in the 5.6 branch is still the only one that fails, so it appears that the Netty version is not the cause of the test failure.

**Comment by jsvd** [comment]
After some investigation about the cert setup failures I also come to the conclusion that the ruby implementation supported pkcs1 certificates (and this is what Flores is generating) while the new implementation expects pkcs8.

it is easy to make the tests pass by using pkcs8 certificates and [netty's SelfSignedCertificates](https://netty.io/4.0/api/io/netty/handler/ssl/util/SelfSignedCertificate.html), but not supporting pkcs1 means a breaking change.

So the missing element at this point is to find a way to accept pkcs1 certificates and pass them correctly to netty. [BouncyCastle seems to be able to handle these certificates](http://www.java2s.com/Tutorial/Java/0490__Security/RSAexamplewithPKCS1Padding.htm), so it is an available last resort.

**Comment by jsvd** [comment]
to maintain backwards compatibility here my idea is to use BC (jruby includes this by default as the underlying software of their openssl implementation) to convert the key to pkcs8:

```
irb(main):117:0> pem_parser = org.bouncycastle.openssl.PEMParser.new(java.io.FileReader.new("/tmp/key"))
=> #<Java::OrgBouncycastleOpenssl::PEMParser:0x5aa6202e>
irb(main):118:0> obj = pem_parser.read_object
=> #<Java::OrgBouncycastleOpenssl::PEMKeyPair:0x771158fb>
irb(main):119:0> kp = org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter.new.get_key_pair(obj)
=> #<Java::JavaSecurity::KeyPair:0x62d0ac62>
irb(main):120:0> kp.private.format
=> "PKCS#8"
```

With a PrivateKey object we can get our SslContextBuilder with [a different `forServer ` call](https://netty.io/4.0/api/io/netty/handler/ssl/SslContextBuilder.html#keyManager-java.security.PrivateKey-java.security.cert.X509Certificate...-)

jsvd:
closing this in favor of #122. This PR's code has been preserved in #122.
The solution for the issues we saw here was to create a fallback builder of SslContext that uses Bouncy castle to load pkcs1 keys and convert them to pkcs8

PR #108, opened 2018-02-01
Since the IP-address it metadata so should port

simmel:
No need to really put the port number into the event. It can be useful
but as metadata.


simmel:
Not sure why the build fails but I don't think it's my code = )

praseodym:
I'd actually suggest that all plugin-added fields should be added to `[@metadata]` so that it never conflicts with fields from the incoming event.

PR #115, opened 2018-03-13
Restores SSLSUBJECT field

### LINKED ISSUE #89: sslsubject field is no longer being set when ssl_verify == true
It looks like https://github.com/logstash-plugins/logstash-input-tcp/commit/e4d4510381b427de8476df82d63961e5031fd532  broke setting the sslsubject field.

The current logstash documentation for the tcp input says when ssl_verify == true,

"Verify the identity of the other end of the SSL connection against the CA. For input, sets the field sslsubject to that of the client certificate."

- Version: logstash 5.6.3

- Operating System: RHEL 7

- Config File (if you have sensitive info, please remove it):
  tcp {
      port =>6514
      tags => ['syslog-tcp-tls']
      ssl_verify => true
      ssl_enable => true
      ssl_cert => '/etc/logstash/logstash_tls.cert'
      ssl_key => '/etc/logstash/logstash_tls.key.pk8'
      ssl_extra_chain_certs => ['/etc/logstash/logstash_tls.ca']
    }

- Steps to Reproduce:  upgrade logstash to 5.6.3 

- Expected behaviour: sslsubject gets set with CN from client certificate

- Actual behaviour: sslsubject does not get set

danhermann:
Fixes #89.


lib/logstash/inputs/tcp.rb
```diff
@@ -291,6 +291,7 @@ def enqueue_decorated(event, client_ip_address, client_address, client_port)
     event.set(HOST_FIELD, client_address) unless event.get(HOST_FIELD)
     event.set(HOST_IP_FIELD, client_ip_address) unless event.get(HOST_IP_FIELD)
     event.set(PORT_FIELD, client_port) unless event.get(PORT_FIELD)
+    event.set(SSLSUBJECT_FIELD, socket.peer_cert.subject.to_s) if @ssl_enable && @ssl_verify && event.get(SSLSUBJECT_FIELD).nil?
```
jsvd: this will crash as there's no `socket` var in the context of this method

PR #116, opened 2018-04-13
New configuration option to allow socket to use TCP keep-alive

jakelandis:


jakelandis:
Here is how i confirmed it is working:

With the following config:
```
input { 
	tcp {  port => 55115      }
}

output { 
   stdout { 
       codec => rubydebug { metadata => true }
      }
} 
```
start logstash

from a different window:
```
irb
require 'socket'
socket = TCPSocket.new("localhost", 55115)
socket.write "hi"
#not closing connection
```

validate the ESTABLISHED state : netstat -an | grep 55115

wait a couple hours and ensure via wireshark that the tcp keep alive packet is sent

jakelandis:
code has been updated with the option, and now applies to SSL connection too.  

I still need to finish testing this which takes bit since I need to hold open a connection for a few hours.

jsvd:
you can test it faster by reducing the keepalive. On osx:

```
sudo sysctl -n net.inet.tcp.keepidle=360000
```

Value is in milliseconds, so 360000 is 10 minutes (default is 7200000, 2 hours)

jakelandis:
@jsvd - can you take another look, I added 2 more commits since last review.

https://github.com/logstash-plugins/logstash-input-tcp/pull/116/commits/e4909897e865c7555d62939a2967fd914884bc18
https://github.com/logstash-plugins/logstash-input-tcp/pull/116/commits/052501a8429fae0008b6efe31554a93536ecc258

I could never get the keep alives to fire from TLS / Ruby side of things. The code (now) is by all accounts is correct, and having burned way to many hours trying to debug, I am calling TLS + keep alive possibly supported ( I think there may be an issue with JRuby).  ... anyway there is PR we should finish up to convert the TLS side of the plugin over to Java, which has many more benifits then just keep alives firing properly when TLS is configured: https://github.com/logstash-plugins/logstash-input-tcp/pull/106

non-TLS (Java/Netty implementation ) tested to work exactly as expected.

jakelandis:
thanks @jsvd !

raymond-grointel:
Can we have a release for 5.0.9? https://rubygems.org/gems/logstash-output-tcp

Issue #110, opened 2018-02-22
Shutdown doesn't close listening port and never stops


simmel:
- Version: 6.2.1
- Operating System: OS X / Linux
- Config File (if you have sensitive info, please remove it):
```
input { tcp { port => "2514" } } output { }
```
- Sample Data:
```
while true; do yes | nc localhost 2514; done
```
- Steps to Reproduce:
```
while true; do yes | nc localhost 2514; done &
bin/logstash -e 'input { tcp { port => "2514" } } output { }'
^C
```

Logstash won't stop until there are no new connections. I.e. I need to stop the while true-loop for Logstash to stop at all. If I don't stop the while true-loop Logstash never shutsdown.

Logs:
When stopping Logstash with ^C/SIGINT:
```
$ bin/logstash -e 'input { tcp { port => "2514" } } output { }'
Sending Logstash's logs to /tmp/logstash-issue/logs which is now configured via log4j2.properties
[2018-02-22T14:27:46,202][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"fb_apache", :directory=>"/tmp/logstash-issue/modules/fb_apache/configuration"}
[2018-02-22T14:27:46,242][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"netflow", :directory=>"/tmp/logstash-issue/modules/netflow/configuration"}
[2018-02-22T14:27:46,668][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2018-02-22T14:27:48,143][INFO ][logstash.runner          ] Starting Logstash {"logstash.version"=>"6.2.1"}
[2018-02-22T14:27:48,910][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2018-02-22T14:27:52,514][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>50}
[2018-02-22T14:28:08,149][INFO ][logstash.inputs.tcp      ] Starting tcp input listener {:address=>"0.0.0.0:2514", :ssl_enable=>"false"}
[2018-02-22T14:28:08,361][INFO ][logstash.pipeline        ] Pipeline started succesfully {:pipeline_id=>"main", :thread=>"#<Thread:0x7e058d5c run>"}
[...]

PR #117, opened 2018-05-29
reorder shut down of the two event loops to prevent shutdown exceptions

jsvd:
fixes https://github.com/logstash-plugins/logstash-input-tcp/issues/110


jsvd:
Thanks @robbavey !

fbaligand:
Great !
@jsvd, could you release it ?

Issue #100, opened 2017-12-15
[5.x.x] Add an option to control the automatic reverse DNS lookup


micoq:
Since the version 5.0.0, a reverse DNS lookup is made for each client connection. Sometimes, this is not desirable.
In pure ruby, `socket.peeraddr[3]` returns by default the peer's IP address if `socket.do_not_reverse_lookup` is `true` (which seem to be the default value on the embedded jruby)
However, the behavior is different in the java loop where `channel_addr.get_host_name` is always called and can do reverse lookups
(see the [getHostName() javadoc](https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html#getHostName()))

It would be nice to have an option to enable or disable the lookups for all cases.

yaauie:
A configurable option was added in version 5.1.0 of this plugin via #121:

~~~ ruby
input {
  tcp {
    # ...
    dns_reverse_lookup_enabled => false
  }
}
~~~

PR #121, opened 2018-08-15
Make DNS Reverse Lookups configurable

yaauie:
Because `InetSocketAddress#getHostName()` can cause the system to do a DNS
reverse lookup which can be undesirable in some scenarios, allow users to
configure the plugin to get the address using an alternate means that does
not incur that penalty.

Resolves: logstash-plugins/logstash-input-tcp#100


Kaacz:
This commit is "The must". 
Forcing DNS Reverse Lookups was BIG mistake. For example in case 200 000+ syslog devices on internet - it is Logstash killer. I was simply remove bad unwanted funcionality, but this is right systematic solution. Thanks.

PR #122, opened 2018-08-31
Port ssl on server mode to netty

### LINKED ISSUE #104: Finish migration to Netty (TLS still uses Ruby Sockets)

jsvd:
replaces #106 

* includes #106 port of ssl server path to netty
* adds a ruby class that creates SslContext if PKCS#1 keys are used
* refactors travis testing setup to use tarballs instead of compiling logstash

If necessary, the travis refactor can be removed as it's contained in a single commit (578969033850bf364de46863bab277052bbd822f)

The initial commit from #106 was preserved.

fixes #104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,9 @@
 ---
 sudo: false
-language: ruby
-cache: bundler
-matrix:
-  include:
-  - rvm: jruby-9.1.13.0
-    env: LOGSTASH_BRANCH=master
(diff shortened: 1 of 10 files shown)


jsvd:
Many thanks for the review @danhermann

PR #124, opened 2018-10-16
simplify ssl handling and fix cert chaining

jsvd:
This pr simplifies the creation of the ssl context by:

- Realizing that the two current strategies only differed in how the private key was loaded, so bouncy castle now is used to produce the PrivateKey on both situations
- Realizing that the two current strategies incorrectly loaded the extra chain certs as trusted certificates to validate the client instead of being used by the client to validate the server, so we now compile a cert+(certificate chain) that is sent to the client.


jsvd:
to make life easier, I added a test for a certificate chain setup w/ verification on both client and server sides. I also added the same test to #125

jsvd:
Many thanks @danhermann!

PR #125, opened 2018-10-19
fix certificate chain setups (backport of #124)

jsvd:
fix is a backport of the fix in #124 for 5.0.x series (which ships with logstash 6.4.x)


jsvd:
Thanks @danhermann

PR #130, opened 2018-11-29
Fix heading for `ssl_certificate_authorities` docs

yaauie:
Doc bug was introduced in #124


hubbleview:
@jsvd  Just would like to check, will this PR backport to the current (6.6) & earlier version of docs in addition to the master branch?

jsvd:
@hubbleview it's not backported by default since the documentation is tied to specific plugin versions that ship with each logstash release.
This fix was added to the 6.0.0 release of this plugin, and logstash 6.x ships with 5.2.0. So my suggestion here is for us to release a 5.2.1 that will be included in the next 6.6.x and 6.5.x releases, what do you think @yaauie ?

PR #138, opened 2019-01-04
[Build] Update dockerfile to use compatible version of Bundler

robbavey:
Bundler 2.0 introduced requirements that are incompatible with the version of Ruby shipped with Logstash 5.6. This commit installs a pre 2.0 version of Bundler. It also removes an irrelevant step from the travis yml


robbavey:
Thanks @danhermann. The build failure predates this commit - https://travis-ci.org/logstash-plugins/logstash-input-tcp/jobs/461685540

PR #140, opened 2019-01-18
support multiple certificates per file

jsvd:
This can be tested easily with the following script: https://gist.github.com/jsvd/b8d21abaf103bdfc7fe8b11008070a95

And then set the certificate chain authorities to the concatenation of the RootCA.crt and IntermediateCA.crt:

`cat RootCA.crt IntermediateCA.crt > chain.crt`

### DIFF
--- a/lib/logstash/inputs/tcp/compat_ssl_options.rb
+++ b/lib/logstash/inputs/tcp/compat_ssl_options.rb
@@ -76,7 +76,9 @@ class SslOptions
     # create certificate object
     cf = CertificateFactory.getInstance("X.509")
     cert_chain = []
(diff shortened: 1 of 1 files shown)


yaauie:
It looks like we have a build failure on 5.x due to a change in behaviour between the activated bouncycastles:

~~~
problem creating RSA private key: java.lang.IllegalArgumentException: failed to construct sequence from byte[]: Extra data detected in stream
~~~

jsvd:
It's a bug in the bouncy castle we ship in 5.6, the fix is in https://github.com/elastic/logstash/pull/10192

yaauie:
Got it. And we only trigger the bug here, in this scenario that is already broken with the current release of the plugin? If so, then I don't see the harm in shipping as-is.

Issue #132, opened 2018-11-30
Error in Netty pipeline: java.io.IOException: Connection reset by peer


damm:
For all general issues, please provide the following details for fast resolution:

- Version: 5.2.0
- Operating System: Ubuntu / Linux
- Config File 
```
  tcp {
    port => '514'
    add_field => { "service" => 'syslog' }
  }
```
- Sample Data:
I don't know if this matters.  It wasn't dropping data.
- Steps to Reproduce:

Configure Logstash to listen on TCP on port 514 
```
  tcp {
    port => '514'
    add_field => { "service" => 'syslog' }
  }
```

- Steps to resolve the problem:

Updating Gemfile.lock to pin it to the `5.1.0` release and then running `logstash-plugin install --version 5.1.0 logstash-input-tcp` 

Lastly, I enabled debug mode and I didn't see anything more.  There is no stack trace of anything crashing in the logs.  It just returns this error for every time a syslog event is received.  No data is lost.

jsvd:
I can't seem to reproduce this:

```
 /tmp/logstash-6.5.1 % bin/logstash -e "input { tcp { port => 3333 } }"
Sending Logstash logs to /tmp/logstash-6.5.1/logs which is now configured via log4j2.properties
[2018-11-30T09:02:11,522][INFO ][logstash.setting.writabledirectory] Creating directory {:setting=>"path.queue", :path=>"/tmp/logstash-6.5.1/data/queue"}
[2018-11-30T09:02:11,531][INFO ][logstash.setting.writabledirectory] Creating directory {:setting=>"path.dead_letter_queue", :path=>"/tmp/logstash-6.5.1/data/dead_letter_queue"}
[2018-11-30T09:02:11,623][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2018-11-30T09:02:11,637][INFO ][logstash.runner          ] Starting Logstash {"logstash.version"=>"6.5.1"}
[2018-11-30T09:02:11,656][INFO ][logstash.agent           ] No persistent UUID file found. Generating new UUID {:uuid=>"6b6225e0-daae-46dd-96d8-876b3df0ed8b", :path=>"/tmp/logstash-6.5.1/data/uuid"}
[...]

PR #142, opened 2019-03-08
fix race condition at startup

jsvd:
currently the socket was being opened during `register` but the queue is only set during `run`, which mean that connections+data could arrive before the a queue was configured, causing the error seen in the linked issue.

This PR changes the socket opening to also happen during `run`.

solves #132


jsvd:
One test wasn't taking into account the move of binding the socket from `#register` to `#run`, fixed now.

PR #150, opened 2020-03-05
Explicitly specify fully qualified package to SslContextBuilder

robbavey:
Since the release of beats 6.0.8, in certain instances
(multiple pipelines where TCP and beats inputs are defined in
separate pipelines), the TCP input is using
org.logstash.netty.SslContextBuilder defined in the beats input instead
of the desired o.netty.handler.ssl.SslContextBuilder from the Netty library.
This causes TCP pipelines to fail to be created, due to a method missing error


robbavey:
Superseded by #152

PR #151, opened 2020-03-12
Build: update gradlew used to latest 4.x

kares:
to fix CI build (HINT: there's still SNAPSHOT failures that need more updates)


kares:
thanks, might merge as is for now and follow up 5.x in another PR. 
as this was just a prequel to cleaning up `java_import`s

PR #152, opened 2020-03-12
Refactor: scope java_import to avoid polluting

kares:
Started at https://github.com/logstash-plugins/logstash-input-beats/pull/388 (having a global constant clash with this plugin). 

The reason looked weird at first but got much cleaner having a simple reproducer: https://github.com/jruby/jruby/issues/6115
Honestly, wasn't sure (wout looking at source) what to expect from an import at the instance-method level. Turns out these issue warnings as they set the same 'global' constant twice (which likely got lost somewhere in LS' logs). 
The real issue is that `java_import` should only have existed on `Module` (and `main`), in which case using it from a i-method would have failed to start with. This is the path JRuby is leaning towards but will take a major release to properly "deprecate" instance imports first.

---

I did review most of the `java_import` usage in plugins and decided to act on a few, esp. ones that require library (non JDK) classes e.g. here `Decoder` sounds a bit "too" common.

NOTE: https://github.com/logstash-plugins/logstash-input-tcp/pull/150 might not be necessary when `SslContextBuilder` constant is scoped at the class level.


robbavey:
Build errors look unrelated

PR #155, opened 2020-04-01
synchronise calls to `Logger#to_java` to prevent potential crash at startup

yaauie:
When a single ruby object receives `RubyObject#to_java` on multiple threads
simultaneously to cast it to a java interface, a concurrency bug in JRuby can
cause a `<NoMethodError: undefined method `__jcreate_meta!' [...]>` exception
to be thrown on one of those threads.

Since this plugin's `@logger` instance variable is a reference to an object
that may be shared with other instances of the plugin, we need to synchronize
sending `LogStash::Logging:Logger#to_java` to ensure that we do not hit the
above concurrency bug.


kares:
for reference, [reported issue upstream](https://github.com/jruby/jruby/issues/6160) as this seems like something JRuby could and should fix

Issue #143, opened 2019-04-10
SSLSUBJECT broken again


markwj:
Was using v5.0.9 for a while, and sslsubject worked well. But, since upgrading to latest (v6.0.2), sslsubject is broken again.

My configuration is:

```
input {
  tcp {
    port => <redacted>
    ssl_enable => true
    ssl_verify => true
    ssl_cert => <redacted>
    ssl_key => <redacted>
    ssl_certificate_authorities => <redacted>
    dns_reverse_lookup_enabled => false
    tcp_keep_alive => true
    codec => json_lines
    id => <redacted>
  }
}
```

A simple

```
output {
  stdout { codec => rubydebug { metadata => true } }
}
```

can show the issue clearly.

I did start some debugging, and find tcp.rb decode_buffer is being called with a nil socket from tcp/decoder_impl.rb decode. Not sure what has changed between v5.0.9 and v6.0.2.

I will start a bisect on this, to try to narrow it down more, and update this issue if I find the cause.

rwaweber:
Seconding this, it appears as though it's still broken as of 6.0.6 with the same pipeline as above.

rwaweber:
I put a PR together internally, needs to go through some internal review first but should be able to submit it in the next couple days hopefully

PR #159, opened 2020-06-17
Fix sslsubject regression

rwaweber:
~- Removes some (what looks like to me) dead code for TLS setup~ Nope! The code was used in client mode, totally misread it.

- Add a regression test to catch if the sslsubject field ever disappears in the future

Attempt to follow the same general flow for tls principal extraction as the beats input plugin:

https://github.com/logstash-plugins/logstash-input-beats/blob/5dd54594f65d32aad87d1dfd7b04d0c801216676/lib/logstash/inputs/beats/message_listener.rb#L125-L155

Extract the ssl subject from inbound messages by:

- adjust the interface to pass in the Netty ChannelHandlerContext instead of an Address object. I think that this opens up the possibility for being able to eventually work in other metadata into TLS-ified objects 

- name the InputHandler's ssl_handler to more reliably retrieve it from the available netty Channel Pipelines.

- in the decoder, check if ssl_verify and ssl_enable are true and pull out the subjectname from the context

Should help close https://github.com/logstash-plugins/logstash-input-tcp/issues/143


rwaweber:
Hey all! Any thoughts on the above?

rwaweber:
Hey all -- friendly poke, would love to get your thoughts on this considering this component is currently broken and seesms to have been that way for quite a while.

rwaweber:
Hey @jsvd -- sorry to ping you directly, but do you have an idea of when this could get looked at?
I'd love to be able to use the `sslsubject` features again in an upcoming release

rwaweber:
Hey Rob! Thanks for the suggestions --  my bad on accidentally snipping the `client` mode chunks, I completely missed those  components.

I'll reinstate the aforementioned methods to not break functionality and incorporate the other suggestions too.

lib/logstash/inputs/tcp/decoder_impl.rb
```diff
@@ -11,24 +11,32 @@ def initialize(codec, tcp)
...
     data.getBytes(0, bytes)
     data.release
     tbuf = String.from_java_bytes bytes, "ASCII-8BIT"
     if @first_read
       tbuf = init_first_read(channel_addr, tbuf)
     end
-    @tcp.decode_buffer(@ip_address, @address, @port, @codec,
-                       @proxy_address, @proxy_port, tbuf, nil)
+    if @tcp.ssl_enable && @tcp.ssl_verify
+      session = ctx.channel().pipeline().get("ssl-handler").engine().getSession()
```
yaauie, line 24: this is a pretty long chain, and it is unclear why we stop here to bind a local variable `session`, which we use only once to continue the chain.

 - are all methods in the chain documented to never return `nil`?
 - should we extract this complexity to a helper method that has appropriate `rescue` clauses?

rwaweber: Hey @yaauie! I stopped there mostly as a means to follow the flow here: https://github.com/logstash-plugins/logstash-input-beats/blob/5dd54594f65d32aad87d1dfd7b04d0c801216676/lib/logstash/inputs/beats/message_listener.rb#L125-L155

I'm happy to add in the rescue that's covered here: https://github.com/logstash-plugins/logstash-input-beats/blob/5dd54594f65d32aad87d1dfd7b04d0c801216676/lib/logstash/inputs/beats/message_listener.rb#L132-L139

rwaweber: > are all methods in the chain documented to never return nil?

It seems that way, aside from the `ChannelPipeline#get` we invoke, though since we explicitly name our channelhandler [here](https://github.com/logstash-plugins/logstash-input-tcp/pull/159/files#diff-ed2ceebf1bfbd6e8dad3c53f7b0c53e0R111), we should be safe in that respect.

>should we extract this complexity to a helper method that has appropriate rescue clauses?

I think thats a great idea, though I sort of think it might be worth tucking in somewhere like the `logstash/base/inputs` package, since we'd be repeating the same logic in this plugin, the beats input plugin, and potentially other plugins where netty is used and TLS metadata extraction is possible. We would need to be careful about formatting though, as different input plugins seem to serialize TLS metadata differently. (it gets cut to 3 fields in the beats plugin and one field here).

Also, as an update on my last message, looking more closely, I'm not sure that rescue clause I cite would really be adding much in terms of safety, but would be more for debug logging insight.

rwaweber:
Hey all! Is there anything else you'd like me to add here?

I think I covered most of @robbavey's earlier concerns, though I'm not quite sure if I've covered @yaauie's points. Happy to continue the discussion.

rwaweber:
Hey all, friendly ping — happy to make some additional changes to get this feature fixed.

rwaweber:
Hey @acchen97 -- any chance you'd be able to lend a hand here?

kervel:
Hello, we are facing the same issue. Any reason why this is not getting merged ?

Issue #146, opened 2019-11-06
Play nicer with ECS


newtonne:
Currently the plugin populates the [host] field with a concrete value. This contradicts ECS which uses a [host object](https://www.elastic.co/guide/en/ecs/current/ecs-host.html), and therefore requires that the host value is renamed before doing anything else if you wish to conform your data to ECS.

I would suggest, instead, populating `[host][hostname]` and/or `[@metadata][host]`.

Thanks.

PR #165, opened 2021-01-18
ECS compatibility

yaauie:
Implements ECS Compatibility mode, supporting `disabled` (legacy), opt-in `v1` (breaking), and an alias from `v8` to `v1`.

In ECS mode, we add tcp connection metadata in a variety of descriptive fields under `[@metadata][tcp]`, instead of the top-level fields that clash with ECS. Users can copy these metadata fields to their events if and when necessary, which allows us to avoid polluting an event with data that may or may-not align with ECS.

---

resolves https://github.com/logstash-plugins/logstash-input-tcp/issues/146


lib/logstash/inputs/tcp/chooser.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  ##
+  # @param choice [Symbol]
+  # @param error_name [#to_s] the name of this choice, to be used in case of an error
+  # @param error_class [Exception] the exception class to use in case of an error
+  # @return [Choice]
+  def choose(choice, error_name="choice", error_class=ArgumentError)
+    if !@supported_choices.include?(choice)
+      message = sprintf("unsupported %s `%s`; expected one of %s", error_name, choice.to_s, @supported_choices.map(&:to_s))
+      logger.error(message)
```
kares, line 23: there seems to be no `logger` method ...

spec/inputs/tcp_spec.rb
```diff
@@ -52,260 +52,293 @@ def get_port
...
-        socket.puts("#{i} ☹")
-        socket.flush
+        insist { events.length } == event_count
+        events = events.sort_by {|e| e.get("message")} # the ordering of events in the queue is highly timing-dependent
+        event_count.times do |i|
+          event = events[i]
+          insist { event.get("message") } == "#{i} ☹"
+
+          case ecs_compatibility
+          when :disabled
```
andsel, line 91: `ecs_compatibility` is used only 2 tests inside the all test cases, and those places are where it's used 
```ruby
case ecs_compatibility
when :disabled ...
when :v1 ...
end
```
I was wondering if:
- the use of `ecs-compatibility` setting has meaning also for the remaining tests?
- the common part of the tests could be extracted in a shared example or some test helper that could share the common part and accept a block to verify the expectations in the 2 cases?
The motivation is that use `if`s and `loop`s in test code make it more unreadable

andsel (commented):
I've left a question on testing part plus a doubt:
- should we also set the `ecs.version` as described in [guidelines](https://www.elastic.co/guide/en/ecs/current/ecs-guidelines.html#_general_guidelines) ?

spec/inputs/tcp_spec.rb
yaauie: I've restructured to (a) use the rspec helpers we published in v1.2 of the support gem, and (2) rearranged so that only those that are liable to have differing behaviour are ent through the variants.

lib/logstash/inputs/tcp/chooser.rb
yaauie: I've extracted the "Chooser" to the ECS Support gem in v1.2, and now this PR uses the shared implementation where we have resolved the problem you mention.

yaauie:
> should we also set the ecs.version as described in guidelines

Unfortunately, we don't have minor/patch versions available to us, so we can't comply :/

lib/logstash/inputs/tcp.rb
```diff
@@ -260,14 +260,24 @@ def handle_socket(socket)
...
+    event.set(@field_host_ip, client_ip_address) unless event.get(@field_host_ip)
+    event.set(@field_port, client_port) unless event.get(@field_port)
+    event.set(@field_sslsubject, socket.peer_cert.subject.to_s) if socket && @ssl_enable && @ssl_verify && event.get(@field_sslsubject).nil?
     decorate(event)
     @output_queue << event
   end
 
+  # setup the field names, with respect to ECS compatibility.
+  def setup_fields!
+    @field_host       = ecs_select[disabled: "host",                    v1: "[@metadata][tcp][source][name]"        ].freeze
```
kares, line 273: what about using `source.address` instead 
so the underlying structure could be copied to a top-level and would be ecs-compliant.

lib/logstash/inputs/tcp.rb
```diff
@@ -260,14 +260,24 @@ def handle_socket(socket)
...
+    @field_host_ip    = ecs_select[disabled: "[@metadata][ip_address]", v1: "[@metadata][tcp][source][ip]"          ].freeze
+    @field_port       = ecs_select[disabled: "port",                    v1: "[@metadata][tcp][source][port]"        ].freeze
+    @field_proxy_host = ecs_select[disabled: "proxy_host",              v1: "[@metadata][tcp][proxy][ip]"           ].freeze
```
kares, line 276: guess we do not have a better (ecs-like) names for `proxy.host` and `proxy.ip` here ...

lib/logstash/inputs/tcp.rb
kares, line 273: in plugins we often use a naming convention for the meta-data e.g. 
- `[@metadata][input][beats][tls][client][subject]`
- or `[@metadata][input][http][request][headers]`

there isn't a tcp filter to conflict here but would still be nice to be explicit using the `[input][tcp]` prefix, wdyt?

we do not follow the convention everywhere , for me personally it would make sense from a user perspective
e.g. having a bare `[@metadata][path]` in file input vs having `[@metadata][input][file][path]`

kares (commented):
:+1: 

do have some suggestion on meta-data field names

spec/inputs/tcp_spec.rb
```diff
@@ -52,160 +54,175 @@ def get_port
...
-        # unicode smiley for testing unicode support!
-        socket.puts("#{i} ☹")
-        socket.flush
+        event_count.times.collect {queue.pop}
       end
-      socket.close
 
-      event_count.times.collect {queue.pop}
-    end
+      insist { events.length } == event_count
```
andsel, line 85: Other `insist` has been moved to `expect`, so maybe also this could be an `expect`

spec/inputs/tcp_spec.rb
```diff
@@ -52,160 +54,175 @@ def get_port
...
 
-    insist { events.length } == event_count
-    events = events.sort_by {|e| e.get("message")} # the ordering of events in the queue is highly timing-dependent
-    event_count.times do |i|
-      insist { events[i].get("message") } == "#{i} ☹"
-      insist { events[i].get("host") } == "1.2.3.4"
-      insist { events[i].get("port") } == "1234"
-      insist { events[i].get("proxy_host") } == "5.6.7.8"
-      insist { events[i].get("proxy_port") } == "5678"
+      insist { events.length } == event_count
```
andsel, line 123: `insist` to `expect`

spec/inputs/tcp_spec.rb
```diff
@@ -52,160 +54,175 @@ def get_port
...
+      socket.puts(text)
       socket.close
 
       queue.pop
     end
 
-    insist { event.get("hello") } == data["hello"]
-    insist { event }.include?("host")
+    # Make sure the 0xA3 latin-1 code converts correctly to UTF-8.
+    insist { event.get("message").size } == 1
```
andsel, line 223: `insist` to `expect`

andsel (approved):
LGTM left only a few nitpicks that you are free to skip

lib/logstash/inputs/tcp.rb
yaauie: There are a few reasons that make wholesale copy of this source metadata impractical.

While `source` and `destination` fields do have an `address`, `ip`, and `port`, they are used to describe network _transactions_ captured _in_ an event. Similarly, `client` and `server` have all three fields, but are used to describe the initiator or recipient of the network transaction.

While the name and ip of the event's source _may_ make sense to copy into `host` or `observer`, the `port` is pretty meaningless and doesn't have a home there. When this plugin is run in server mode, the outbound port from the client is typically chosen at random by its host OS, and when the plugin is run in client mode, the port we are connecting to doesn't really describe the _event_.

I chose `name` here because it aligns with both `host` and `observer`.

We _could_ align with the CEF codec, which has a new `device` option that accepts `host` or `observer`, and could populate the relevant bits if this is set?

kares: thanks for sharing the details - fine by me to keep as is - prefixed `source.name`

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 6.2.0
+ - Added ECS Compatibility Mode [#165](https://github.com/logstash-plugins/logstash-input-tcp/pull/165)
+   - When operating in an ECS Compatibility mode, metadata about the connection on which we are receiving data is nested in well-named fields under `[@metadata][tcp]` instead of at the root level.
```
kares, line 3:
```suggestion
   - When operating in an ECS Compatibility mode, metadata about the connection on which we are receiving data is nested in well-named fields under `[@metadata][input][tcp]` instead of at the root level.
```

kares (approved):
:icecream:

docs/index.asciidoc
```diff
@@ -108,6 +119,58 @@ It is possible to avoid DNS reverse-lookups by disabling this setting. If disabl
...
+** `disabled`: unstructured connection metadata added at root level
+** `v1`,`v8`: structured connection metadata added under `[@metadata][input][tcp]`
+* Default value depends on which version of Logstash is running:
+** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+** Otherwise, the default value is `disabled`.
+
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)].
+The value of this setting affects the placement of a TCP connection's metadata on events:
+
+.Source Metadata Location by `ecs_compatibility` value
```
karenzone, line 136:
```suggestion
The value of this setting affects the placement of a TCP connection's metadata on events.

```

docs/index.asciidoc
```diff
@@ -108,6 +119,58 @@ It is possible to avoid DNS reverse-lookups by disabling this setting. If disabl
...
+
+.1+|SSL Subject Metadata from a secured TCP
+    connection. Available when `ssl_enable => true`
+    AND `ssl_verify => true`                        l|[@metadata][input][tcp][ssl][subject] l|[sslsubject]
+|=======================================================================
+
+For example, the Elastic Common Schema reserves the https://www.elastic.co/guide/en/ecs/current/ecs-host.html[top-level `host` field] for information about the host on which the event happened.
+If an event is missing this metadata, it can be copied into place from the source TCP connection metadata that has been added to the event:
+
+[source]
```
karenzone, line 160:
```suggestion
[source,txt]
```

docs/index.asciidoc
karenzone: On other plugins we've standardized putting the metadata table in the ECS metadata section.  There's a compelling argument for either location. If you don't have strong feelings, perhaps we could move this under the ECS metadata section.

karenzone: Nice job on the table. It renders quite nicely.

karenzone (commented):
Suggestions inline for your consideration.

docs/index.asciidoc
yaauie: I've moved the table, and re-worded the relevant links/intros.

docs/index.asciidoc
```diff
@@ -70,6 +70,53 @@ event timestamp
...
+[id="plugins-{type}s-{plugin}-ecs_metadata"]
+==== Event Metadata and the Elastic Common Schema (ECS)
+
+In addition to decoding the events, this input will add metadata about the TCP connection itself to each event.
+This can be helpful when applications are configured to send events directly to this input's TCP listener without including information about themselves.
+
+Historically, this metadata was added to a variety of non-standard top-level fields, which had the potential to create confusion and schema conflicts downstream.
+With ECS compatibility mode, we can ensure a pipeline still has access to this metadata throughout the event's lifecycle without polluting the top-level namespace.
+
+.Source Metadata Location by `ecs_compatibility` value
```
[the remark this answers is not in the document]
karenzone: We can remove the table title. We don't use numbered tables, and the info in the title is redundant.

karenzone (approved):
Suggestion to removed numbered table title. Otherwise, LGTM! 🎉

PR #169, opened 2021-03-18
[Doc] Better wording for 'tcp_keep_alive' option

andsel:
Describes better what's the meaning of the `tcp_keep_Alive` setting


docs/index.asciidoc
```diff
@@ -202,7 +202,9 @@ For input, sets the field `sslsubject` to that of the client certificate.
   * Value type is <<boolean,boolean>>
   * Default value is `false`
 
-Instruct the socket to use TCP keep alives. Uses OS defaults for keep alive settings.
+Instruct the socket to use TCP keep alive. If it's `true` then the underlying socket
+will use the OS defaults settings for keep alive. If it's `false` it doesn't configure any
+keep alive setting for the underlying socket.
```
karenzone, line 207:
```suggestion
Instruct the socket to use TCP keep alive. If `true` the underlying socket
will use the OS default settings for keep alive. If `false` no
keep alive settings are configured for the underlying socket.
```

karenzone (commented):
Thanks for adding this clarification, @andsel.  I made some minor suggestions for your consideration.

karenzone (approved):
LGTM. Thanks for adding this clarification

karenzone:
Published as v.6.0.8:  https://rubygems.org/gems/logstash-input-tcp/versions/6.0.8-java

PR #170, opened 2021-03-18
DOCTEST: Add a second file in order to test impact

karenzone:
Testing to see if we can safely add a second file to the `/docs/` directory without disrupting normal docgen processing. 

Goal: Find the best approach for handling aliased plugins.

Test Plan: 
* Introduce test file (do not include in changelog).
* Bump version and publish gem.
* Generate plugin docs to be sure that extra file doesn't cause problems. Look for any unexpected consequences. 
* Delete testfile. 

@Andsel Perhaps we can work together to run this test when you finalize #169. This could be relevant to other project we're working on together.

### DIFF
--- a/docs/testfile.asciidoc
+++ b/docs/testfile.asciidoc
@@ -1,0 +1,30 @@
+:plugin: tcp-test
+:type: input
+:default_codec: line
+
+///////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


andsel (approved):
LGTM

Issue #173, opened 2021-04-21
Our version of commons-io is a bit old


joshbressers:
The version of commons-io that we include in the logstash-input-tcp is a bit on the old side and needs an update.

https://github.com/logstash-plugins/logstash-input-tcp/blob/1472a23b6c97b6cd0256453cf82693bba62fbc67/build.gradle#L56

We include version 2.5 which is from 2016. The latest version at the time of this writing is 2.8

PR #174, opened 2021-04-22
bump common-io version

kaisecheng:
This PR update the old common-io lib

Fix https://github.com/logstash-plugins/logstash-input-tcp/issues/173


build.gradle
```diff
@@ -53,7 +53,7 @@ shadowJar {
 dependencies {
   compile 'io.netty:netty-all:4.1.49.Final'
 
-  compile group: 'commons-io', name: 'commons-io', version: '2.5'
+  compile group: 'commons-io', name: 'commons-io', version: '2.8'
```
jsvd, line 56:
```suggestion
  compile group: 'commons-io', name: 'commons-io', version: '2.8.0'
```

jsvd (commented):
minor tweak since commons-io 2.8 version is actually 2.8.0
before merging please update the changelog and bump the version as well

jsvd (approved):
LGTM

Issue #177, opened 2021-06-15
improve error logging to always reveal SSL details


kares:
due [Netty SSLException wrapping](https://github.com/netty/netty/blob/netty-4.1.49.Final/handler/src/main/java/io/netty/handler/ssl/JdkSslServerContext.java#L288) plugin can easily fail with a wrapped error without any additional logging on the wrapped cause (Logstash currently does not log causes when handling errors at the Ruby level), sample logged exception:

[ERROR][logstash.javapipeline    ][main] Pipeline error {:pipeline_id=>"main", :exception=>javax.net.ssl.SSLException: failed to initialize the server-side SSL context, :backtrace=>["io.netty.handler.ssl.JdkSslServerContext.newSSLContext(io/netty/handler/ssl/JdkSslServerContext.java:288)", "io.netty.handler.ssl.JdkSslServerContext.<init>(io/netty/handler/ssl/JdkSslServerContext.java:247)", "io.netty.handler.ssl.SslContext.newServerContextInternal(io/netty/handler/ssl/SslContext.java:465)", "io.netty.handler.ssl.SslContextBuilder.build(io/netty/handler/ssl/SslContextBuilder.java:571)", "jdk.internal.reflect.GeneratedMethodAccessor263.invoke(jdk/internal/reflect/GeneratedMethodAccessor263)", "jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(jdk/internal/reflect/DelegatingMethodAccessorImpl.java:43)", "java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:566)", "org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(org/jruby/javasupport/JavaMethod.java:441)", "org.jruby.javasupport.JavaMethod.invokeDirect(org/jruby/javasupport/JavaMethod.java:305)", "usr.share.logstash.vendor.bundle.jruby.$2_dot_5_dot_0.gems.logstash_minus_input_minus_tcp_minus_6_dot_0_dot_7_minus_java.lib.logstash.inputs.tcp.compat_ssl_options.toSslContext(/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-input-tcp-6.0.7-java/lib/logstash/inputs/tcp/compat_ssl_options.rb:127)",

PR #178, opened 2021-06-15
Feat: improve SSL error logging/unwrapping

kares:
To be able to easily log full Java SSL exception details, we first move the SSL builder to Java.

otherwise this is ~ same as https://github.com/logstash-plugins/logstash-input-beats/pull/405 : 
> The unwrapping at the Java level is for [exceptions wrapped by Netty](https://github.com/netty/netty/blob/netty-4.1.49.Final/handler/src/main/java/io/netty/handler/ssl/JdkSslServerContext.java#L288).

> Full exception details will be logged at debug level from the Java side - since we seem to prefer (manual) exception logging at the plugin level. We also make sure to log cause, if any, on the Ruby side which now catches all (expected) Java exceptions.

**NOTE:** we also avoid installing the bouncy-castle security provider at runtime, [done previously](https://github.com/logstash-plugins/logstash-input-tcp/blob/v6.0.10/lib/logstash/inputs/tcp/compat_ssl_options.rb#L87)

resolves https://github.com/logstash-plugins/logstash-input-tcp/issues/177


src/main/java/org/logstash/tcp/SslContextBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+        final ArrayList<Certificate> certificates = new ArrayList<>();
+        final FileInputStream fis = new FileInputStream(file);
+        try {
+            while (fis.available() > 0) {
+                try {
+                    certificates.add( factory.generateCertificate(fis) );
+                } catch (CertificateException ex) {
+                    Throwable cause = ex.getCause();
+                    if (cause != null && "Empty input".equals(cause.getMessage())) {
+                        continue;
```
robbavey, line 175: Is there a case for `debug` or `trace` logging upon receiving this error, or is it always benign?

robbavey (approved):
minor debug logging question, otherwise LGTM

src/main/java/org/logstash/tcp/SslContextBuilder.java
kares: good call, I simply converted from the .rb parts and did not revisit with potential logging.
the whole logic seems a bit unusual - would rather just use `generateCertificates` to read multiple entries.
anyhow, let's keep that as before and I am going to add a debug line around the caught "empty" exception.

kares: added debug logging + an extra check when expecting to read at least one certificate from `ssl_cert`
also worth noting (in the changelog) we no longer have the side effect of setting up the BC provider globally.

robbavey (approved):
LGTM

PR #181, opened 2021-08-25
Fix: restore BC provider installation

kares:
With the Ruby -> Java rewrite in https://github.com/logstash-plugins/logstash-input-tcp/pull/178 we avoided installing the BC provider at runtime.
Unfortunately to be able to decrypt an openssl key pair the BC class relies on the fact that the provider is available.

Otherwise reading the key-pair leads to a `NoSuchAlgorithmException: PBKDF-OpenSSL SecretKeyFactory not available`


robbavey (approved):
LGTM!

Thank you so much for adding all these test cases!

PR #183, opened 2021-10-17
Syntax errors in example

lrlnt:
There are couple of syntax mistakes in example. "not" expression needs "()" around "[host]" variable and after "copy" must be "=>".

filter {
  if [@metadata][input][tcp][source] and not ([host]) {
    mutate {
      copy => {
        "[@metadata][input][tcp][source][name]" => "[host][name]"
        "[@metadata][input][tcp][source][ip]"   => "[host][ip]"
      }
    }
  }
}

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -105,9 +105,9 @@ If an event is missing this metadata, it can be copied into place from the sourc
 [source,txt]
 -----
 filter {
-  if [@metadata][input][tcp][source] and not [host] {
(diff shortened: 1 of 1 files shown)


andsel:
Hi @lrlnt your suggestion was good, but that code sample has been moved to https://github.com/logstash-plugins/logstash-input-tcp/blob/main/docs/index.asciidoc#L108, so I close this PR since not anymore actual.

PR #185, opened 2021-11-10
Deps: relax jruby-openssl upper bound

kares:
An improvement to avoid running into upgrading issues in Logstash itself, in the future.

let Logstash core dictate what version of jruby-openssl is available, 
the lower bound >= requirement is sufficient for the plugin's needs


andsel (approved):
LGTM

PR #187, opened 2021-12-15
Build: skip shadowing jar dependencies

kares:
There's 2 changes here (they can be split into 2 PRs if needed): 

### 1. avoid shadow-ing dependencies

the shadow repackages dependencies into the `logstash-input-tcp-X.Y.Z.jar`, meaning e.g. classes from `netty-all-4.1.65.Final.jar` will all end up as part of the plugin jar on `$CLASSPATH`. this is bad and if we would to have any kind of feature detection based on a class present would lead to issues:

here's what happens when 2 plugins load the same dependency (one shadowed):
- `require_jar('io.netty', 'netty-all', '4.1.65.Final')` and `require_jar('org.logstash.plugins.input.http', 'logstash-input-http', '3.4.2')` we end up with 2 jars on the `$CLASSPATH`
- `require_jar('org.logstash.inputs', 'logstash-input-tcp', '6.2.5')` adds one jar to the `$CLASSPATH` that contains either the same or a different version of classes already loaded (depending on whether Netty version is the same)

after shadowing removed here's what happens:

- `require_jar('io.netty', 'netty-all', '4.1.65.Final')` and `require_jar('org.logstash.plugins.input.http', 'logstash-input-http', '3.4.2')` we end up with 2 jars on the `$CLASSPATH`
- `require_jar('io.netty', 'netty-all', '4.1.65.Final')` and `require_jar('org.logstash.inputs', 'logstash-input-tcp', '6.2.5')` will only add the later to the `$CLASSPATH` (even if netty's version was different jar-dependencies only loads one dependency with the same name)

### 2. make log4j-api a 'provided' dependency

the **log4j-api** should be considered a transitive dependency of `logstash-core`. 
we need the `log4j-api` jar for compiling but should not package the .jar into the plugin - LS lacks abstraction of custom logger impl for Java plugins thus the log4j-api should simply be considered a provided dependency.


lib/logstash-input-tcp_jars.rb
```diff
@@ -1,4 +1,7 @@
-# encoding: utf-8
+# AUTOGENERATED BY THE GRADLE SCRIPT. DO NOT EDIT.
```
kares, line 1: the generated file no longer contains `require_jar('org.apache.logging.log4j', 'log4j-api', '2.17.0')`
the file could be .gitignore-d like other plugins do.

build.gradle
```diff
@@ -44,29 +44,46 @@ repositories {
   mavenCentral()
 }
 
-apply plugin: 'com.github.johnrengelman.shadow'
```
jsvd, line 47: I believe we can also remove the dependency on the shadow plugin on line 38-40

build.gradle
```diff
@@ -44,29 +44,46 @@ repositories {
...
-
 dependencies {
   testImplementation 'org.apache.logging.log4j:log4j-core:2.17.1'
 
   implementation 'io.netty:netty-all:4.1.65.Final'
   implementation group: 'commons-io', name: 'commons-io', version: '2.8.0' // used from Netty
-  implementation 'org.apache.logging.log4j:log4j-api:2.17.1'
-
+  compileOnly 'org.apache.logging.log4j:log4j-api:2.17.1' // provided by Logstash
+  compileOnly group: 'org.jruby', name: 'jruby-complete', version: "9.1.13.0" // provided by Logstash
```
jsvd, line 53: No need to bump netty and bouncy castle and versions in this PR but this is a new line:

```suggestion
  compileOnly group: 'org.jruby', name: 'jruby-complete', version: "9.2.20.1" // provided by Logstash
```

jsvd (approved):
minor suggestions but otherwise LGTM.

build.gradle
kares: going to leave it as is for now - there isn't a `required_ruby_version >= ...` thus plugin might still work on old LS

PR #198, opened 2022-02-03
Feat: ssl_supported_protocols (TLSv1.3) + ssl_cipher_suites

kares:
The intent here is to officially support TLS 1.3 with an option to limit the protocol (both in server/client mode) with:
- `ssl_supported_protocols => TLSv1.3`

As well as being able to limit the TLS cipher suites used between client and server using:
- `ssl_cipher_suites => [ ... ]`

Unfortunately to be able to support the feature a recent version of JRuby-OpenSSL is required and since the gem version is locked in 7.x and 8.0 versions we need to play some clever tricks to disable the `>= 0.12.2` requirement -> tests than feature detect the version available and adjust accordingly.


docs/index.asciidoc
```diff
@@ -224,6 +226,19 @@ to the connecting clients.
...
...TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256, TLS_DHE_RSA_WITH_AES_256_CBC_SHA256, TLS_DHE_DSS_WITH_AES_256_CBC_SHA256, TLS_DHE_RSA_WITH_AES_128_CBC_SHA256, TLS_DHE_DSS_WITH_AES_128_CBC_SHA256, TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA384, TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384, TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256, TLS_ECDH_RSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA, TLS_DHE_RSA_WITH_AES_256_CBC_SHA, TLS_DHE_DSS_WITH_AES_256_CBC_SHA, TLS_DHE_RSA_WITH_AES_128_CBC_SHA, TLS_DHE_DSS_WITH_AES_128_CBC_SHA, TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA, TLS_ECDH_RSA_WITH_AES_256_CBC_SHA, TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA, TLS_ECDH_RSA_WITH_AES_128_CBC_SHA, TLS_RSA_WITH_AES_256_GCM_SHA384, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_256_CBC_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA256, TLS_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_CBC_SHA ]`
```
kares, line 237: we could try defaulting to [Mozilla's](https://wiki.mozilla.org/Security/Server_Side_TLS) intermediate cipher set but unlike with Beats we do not control both ends of the stack, thus anything diverging from the default set supported by Java could be considered a breaking change, right?

lib/logstash/inputs/tcp.rb
```diff
@@ -112,6 +112,13 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
   # All the certificates will be read and added to the trust store.
   config :ssl_certificate_authorities, :validate => :array, :default => []
 
+  # NOTE: the default setting [] uses Java SSL engine defaults.
+  config :ssl_supported_protocols, :validate => ['TLSv1.1', 'TLSv1.2', 'TLSv1.3'], :default => [], :list => true
```
kares, line 116: ES (still) supports legacy protocols as well I believe it would make sense to support `SSLv3` and `TLSv1` as well

kares:
```suggestion
  config :ssl_supported_protocols, :validate => ['SSLv3', 'TLSv1', 'TLSv1.1', 'TLSv1.2', 'TLSv1.3'], :default => [], :list => true
```

yaauie (commented):
Partial review, will come back for more.

spec/inputs/tcp_spec.rb
```diff
@@ -646,6 +650,75 @@ def get_port
...
+              expect(used_tls_protocol).to eql tls_version
+            end
+          end
+
+          context "with enforced protocol range" do
+            let(:config) do
+              base_config.merge 'ssl_supported_protocols' => [ 'TLSv1.3', 'TLSv1.2' ]
+            end
+            let(:sslcontext) do
+              super().tap { |ctx| ctx.ssl_version = 'TLSv1.2' }
```
yaauie, line 682: comment for clarity: I believe this `sslcontext` is our _external_ SSL that will be establishing a connection to our input plugin, and that we are setting its ssl_version to something that is _included_ in the range supported by our input?

If so, I would hope for another spec, where our external SSL specifies a version that cannot be satisfied by our input, even though that would be largely spec-ing the error as apparent to the external, because we want to make sure that a mismatch doesn't crash our plugin.

spec/inputs/tcp_spec.rb
```diff
@@ -646,6 +650,75 @@ def get_port
...
+              expect(used_tls_protocol).to eql 'TLSv1.2'
+            end
+          end
+
+          context "with specified cipher suites" do
+            let(:config) do
+              base_config.merge 'ssl_cipher_suites' => [ cipher_suite ]
+            end
+
+            let(:cipher_suite) { 'TLS_RSA_WITH_AES_128_GCM_SHA256' }
```
yaauie, line 705: similarly here, I would like a spec for behaviour when negotiation for a common cipher suite fails to ensure that our plugin handles it gracefully without crashing.

spec/inputs/tcp_spec.rb
```diff
@@ -717,4 +790,44 @@ def get_port
...
+      end
+
+      let(:cipher_suite) { 'TLS_RSA_WITH_AES_128_GCM_SHA256' }
+
+      it "sets ciphers" do
+        cipher_ary = ssl_context.ciphers.first
+        expect( cipher_ary[0] ).to eql 'AES128-GCM-SHA256'
+      end
+
+    end if JOpenSSL::VERSION >= '0.12.2'
```
yaauie, line 829: nitpick: I find it very easy to miss the conditional in `context do`...`end if <condition>` blocks.

lib/logstash/inputs/tcp.rb
```diff
@@ -112,6 +112,13 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
...
+
+  # The list of ciphers suite to use, listed by priorities.
+  # NOTE: the default setting [] uses Java SSL defaults.
+  config :ssl_cipher_suites, :validate => :array, :default => []
```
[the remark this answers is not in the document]
yaauie: Perhaps `getSupportedCipherSuites` defined in this PR?

docs/index.asciidoc
yaauie: Is it possible to link to external documentation, or to hide this behind a click? This is a lot of specifics and I worry that the signal will be lost in the noise.

Alternatively, is it possible to query for these using the `bin/ruby` executable or some other means so that a user can query them on their own specific system?

yaauie (commented):
Over-all implementation makes sense. I've left a few nitpicks.

My primary concern is the altered code-path that we take in CI to _not_ specify a dependency on JOSSL.

docs/index.asciidoc
```diff
@@ -258,6 +273,23 @@ The path to the private key corresponding to the specified certificate (PEM form
...
+
+  * Value type is <<string,string>>
+  * Allowed values are: `'TLSv1.1'`, `'TLSv1.2'`, `'TLSv1.3'`
+  * Default depends on the JDK being used. With up-to-date Logstash, the default is `['TLSv1.2', 'TLSv1.3']`.
+    `'TLSv1.1'` is not considered secure and is only provided for legacy applications.
+
+List of allowed SSL/TLS versions to use when establishing a secure connection.
+
+For Java 8 `'TLSv1.3'` is supported  only since **8u262** (AdoptOpenJDK), but requires that you set the
+`LS_JAVA_OPTS="-Djdk.tls.client.protocols=TLSv1.3"` system property in Logstash.
```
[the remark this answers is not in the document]
kares: add this point - this docs part is a copy-pasta from other plugins: https://github.com/logstash-plugins/logstash-output-http/blob/v5.5.0/docs/index.asciidoc#plugins-outputs-http-ssl_supported_protocols=

logstash-input-tcp.gemspec
```diff
@@ -23,10 +23,16 @@ Gem::Specification.new do |s|
...
 
-  s.add_runtime_dependency 'logstash-core', '>= 6.7.0'
-
-  # we depend on bouncycastle's bcpkix-jdk15on being on the class-path
-  s.add_runtime_dependency 'jruby-openssl', '>= 0.10.2'
+  s.add_runtime_dependency 'logstash-core', '>= 6.7.0' # LS 6.7: JRuby 9.2.7.0
+
+  # NOTE: be able to test against LS 7.x / 8.x
+  # (some LS versions have strict jruby-openssl requirements in logstash-core e.g. "= 0.11.0")
+  if ENV['ELASTIC_CONTAINER'].eql?('true') && ENV['PWD'].eql?('/usr/share/plugins/plugin')
```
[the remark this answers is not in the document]
kares: > So under CI we simply don't declare a runtime dependency on JOSSL?

correct we assume the one from logstash-core to be used (which is either smt limited e.g. `= 0.11.0` or latest due smt like `~> 0.11.0`).

> How do we ensure that we don't build a gem artifact that is missing this dependency declaration?

we assume the gem is not to be build in a docker container used for testing (plugin at `/usr/share/plugins/plugin`). I am open to better suggestions - this is an ugly compromise until we unlock jossl version in LS 7.x / 8.x (pending)

spec/inputs/tcp_spec.rb
kares: that's a bit of a redundant spec (pretty much testing the underlying SSL layer) but added as requested ...

docs/index.asciidoc
kares: rendering wise, [Beats listing cipher suites](https://www.elastic.co/guide/en/logstash/master/plugins-inputs-beats.html#plugins-inputs-beats-cipher_suites) is similar (although shorter) - lines just break.

anyhow, updated to use a `bin/ruby -e ...` script as suggested ...

docs/index.asciidoc
```diff
@@ -224,6 +226,18 @@ to the connecting clients.
 Validate client certificate or certificate chain against these authorities.
 You can define multiple files or paths. All the certificates will be read and added to the trust store.
 
+[id="plugins-{type}s-{plugin}-ssl_cipher_suites"]
+===== `ssl_cipher_suites`
+
+  * Value type is <<string,string>>
+  * Default value depends on JDK defaults
```
yaauie, line 233:
```suggestion
  * Default value includes _all_ cipher suites enabled by the JDK and depends on JDK configuration
```

yaauie (commented):
The changes themselves make sense, and docs are getting _"close enough"_.

My only real hold-out is the JOSSL dependency declaration. As is, the built gem artifact will declare a dependency on JOSSL that _cannot_ be satisfied on older Logstashes (prohibiting it from being installed), even though our CI validates that it works on those older Logstashes and the JOSSL that they provide. If the plugin works on those older versions we should not constrain it unnecessarily.

logstash-input-tcp.gemspec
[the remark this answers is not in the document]
kares: we can not provide all the features we ship here wout the latest JOSSL, tests simply work-around and not test the features when up-to-date JOSSL isn't available (this is to change later when old LS versions aren't locked to a specific JOSSL version - e.g. we recently unlocked 7.17 to not limit JOSSL to `= 0.11.1`).

- in client mode TLSv1.3 is only supported since JOSSL 0.12
- the ability to set Java cipher names on the Ruby SSLContext (client mode) only works since 0.12.2
- installing latest tcp plugin version is expected to work on 7.x (when jossl isn't strictly locked) as well as 8.x moving forward (in 8.x again the strict lock has been relaxed back)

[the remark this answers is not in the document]
kares: the JOSSL features are mentioned through the code as comments:
- https://github.com/logstash-plugins/logstash-input-tcp/pull/198/files#diff-012c902c5ac82e3d122ada84ad7e711f565ff0187dabc4cca2e3a56215ba05b3R313
- https://github.com/logstash-plugins/logstash-input-tcp/pull/198/files#diff-48a0ca35b36296ce60b56aadd06fdbee9dfd628a74e0a69138157d430ca78cb1R34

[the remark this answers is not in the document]
yaauie: > we can not provide all the features we ship here wout the latest JOSSL,

This makes sense.

> tests simply work-around and not test the features when up-to-date JOSSL isn't available 

If the distributable artifact of this plugin cannot be _installed_ on a Logstash whose JOSSL isn't latest (because of pinning in Logstash core), why are we validating a subset of this plugin's behaviour in the presence of a potentially-older JOSSL at all? Should we instead gatekeep the _feature_ and warn/error when it is invoked without the prerequisite met?

***

If a subset of the functionality introduced in this PR is not available on older Logstashes (that have JOSSL pinned), and we include a dependency declaration in our gem artifact that prevents them from being installed on those older Logstashes, why are we going out of our way in CI to ensure that the rest of the functionality works? None of the functionality will work because the plugin manager will refuse to install it.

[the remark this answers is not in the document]
kares: so your proposal is to e.g. have a :red_circle: 7.x built (at least until 7.17.2 is released which unlocks jossl) ?
when this feature started 8.x has locked jossl as well.

the build is going to fail in the image build step (during `bundle install`)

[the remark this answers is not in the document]
yaauie: It is pointless to have green CI for a version of Logstash upon which the distributable version of the artifact cannot be installed. 

If some subset of the functionality is expected to work with older JOSSL, then newest JOSSL isn't actually a runtime requirement of the whole.

My proposal is that we specify the minimum JOSSL we _actually_ require, and reject configuration _at runtime_ when that configuration requires a newer JOSSL than we have. For example, if specifying ciphers requires a newer JOSSL than is available at runtime, we raise a config error if and only if the user specifies ciphers. Similarly, if TLSv1.3 requires a newer JOSSL than we have, we raise a config error if and only if the user specifies TLSv1.3.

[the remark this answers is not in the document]
kares: > It is pointless to have green CI for a version of Logstash upon which the distributable version of the artifact cannot be installed.

as noted it was meant to be a "temporary measure" - to me it made sense to have a :green_circle: than for others to get confused why CI is failing for 7.x ... anyhow it's now :red_circle: as I do not care about my preference here if a :red_circle: CI is acceptable (as I forced the `>= 0.12.2` requirement).

going to look into setting up some older 7.x version manually but it does not feel like time well spent ...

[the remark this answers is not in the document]
kares: added extra runtime check on Java 8 for when `TLSv1.3` would not work I am not 100% it's worth the effort given Java 8 is effectively dead - the assumption I operated with was that documenting this is more than enough ... also other plugins, already shipped with the new `ssl_supported_protocols` setting, do not have this extra check atm.

logstash-input-tcp.gemspec
```diff
@@ -23,10 +23,10 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "logstash-core-plugin-api", ">= 1.60", "<= 2.99"
   s.add_runtime_dependency 'logstash-mixin-ecs_compatibility_support', '~>1.2'
 
-  s.add_runtime_dependency 'logstash-core', '>= 6.7.0'
+  s.add_runtime_dependency 'logstash-core', '>= 6.7.0' # LS 6.7: JRuby 9.2.7.0
 
   # we depend on bouncycastle's bcpkix-jdk15on being on the class-path
-  s.add_runtime_dependency 'jruby-openssl', '>= 0.10.2'
+  s.add_runtime_dependency 'jruby-openssl', '>= 0.12.2' # 0.12 supports TLSv1.3
```
yaauie, line 29: This change is effectively breaking, preventing this minor-bumped feature from being installed on _most_ Logstashes in the 7.x series where JOSSL is pinned, despite this PR otherwise doing a fair bit of work to ensure that the old functionality could run on those older Logstashes.

Can we continue to pin to `>= 0.10.2`, which _will_ select JOSSL 0.12 on newer Logstashes and will allow installation on older Logstashes where the upgraded JOSSL is not available? If so, to _prove_ that we can run the subset of behaviours we test for, can we include an old-JOSSL-pinned Logstash in the matrix (say: 7.17.2, [which is pinned to `~> 0.11`](https://github.com/elastic/logstash/blob/v7.16.2/Gemfile.jruby-2.5.lock.release#L20))

kares:
@yaauie the plugin is updated to require LS `>= 8.1` (which already [has](https://github.com/elastic/logstash/blob/v8.1.3/Gemfile.jruby-2.5.lock.release#L172=) jruby-openssl 0.12.2)
this means we no longer need special checks for Java 8 (due TLSv1.3 support) thus all of the code has been removed.

lib/logstash/inputs/tcp.rb
```diff
@@ -297,6 +304,14 @@ def ssl_context
         @ssl_context.cert_store  = load_cert_store
         @ssl_context.verify_mode = OpenSSL::SSL::VERIFY_PEER|OpenSSL::SSL::VERIFY_FAIL_IF_NO_PEER_CERT
       end
+      if ssl_supported_protocols.any?
+        min_max_version = ssl_supported_protocols.sort.map { |n| n.sub('v', '').sub('.', '_').to_sym } # 'TLSv1.2' => :TLS1_2
+        @ssl_context.min_version = min_max_version.first
+        @ssl_context.max_version = min_max_version.last
```
jsvd, line 310: as mentioned in the tcp output review, we can enable the exact list of supported protocols in the context instead of max/min, so that the behaviour is more consistent with the java context.

AnandhkumarG:
Hi , Does this feature require Logstash 8.x as minimum? WIll this feature work in 7.1.7.x for example. Please let know if there is a compatiibility matrix for plugins versions-LS versions. Thanks in advance

PR #199, opened 2022-03-29
Fix SSL Subject regression in server mode

yaauie:
Alters the TCP#decode_buffer signature to accept the _ssl subject_ instead of
expecting a ruby socket, so that it can be interoperable between the ruby-based
client mode and the netty-powered server mode.

In server mode, the SSL subject is extracted _once_ when initializing the
connection IFF SSL is enabled and verification is turned on.

Co-authored-by: Will Weber <rwa_weber@comcast.com>
Closes: logstash-plugins/logstash-input-tcp#159


lib/logstash/inputs/tcp/decoder_impl.rb
```diff
@@ -53,10 +55,20 @@ def init_first_read(channel_addr, received)
...
+    @sslsubject = extract_sslsubject(channel)
     @first_read = false
     filtered
   end
 
+  private
+  def extract_sslsubject(channel)
+    return nil unless @tcp.ssl_enable && @tcp.ssl_verify
+
+    channel.pipeline().get("ssl-handler").engine().getSession().getPeerPrincipal().getName()
```
kares, line 67:
```suggestion
    channel.pipeline().get("ssl-handler").engine.getSession.getPeerCertificates[0].getSubjectX500Principal.getName
```

lib/logstash/inputs/tcp/decoder_impl.rb
```diff
@@ -53,10 +55,20 @@ def init_first_read(channel_addr, received)
...
+  rescue Exception => e
```
kares, line 68: let's not use the anti-pattern or rescuing everything
```suggestion
  rescue java.lang.Exception
```
also might be worth logging the exception at least on the debug level

lib/logstash/inputs/tcp.rb
```diff
@@ -215,10 +215,14 @@ def run_client()
...
   end
 
+  # only called in client mode
   def handle_socket(socket)
     client_address = socket.peeraddr[3]
     client_ip_address = socket.peeraddr[2]
     client_port = socket.peeraddr[1]
+
+    # Client mode sslsubject extraction, server mode happens in DecoderImpl#decode
+    ssl_subject = socket.peer_cert.subject.to_s if @ssl_enable && @ssl_verify
```
kares, line 225:
```suggestion
    ssl_subject = socket.peer_cert.subject.to_s(OpenSSL::X509::Name::RFC2253) if @ssl_enable && @ssl_verify
```

kares (commented):
not sure what the old format was but it would be nice to use the same (RFC) format.
otherwise the formats are different (the RFC format is cleaner IMHO) : 
- Ruby `"/OU=foo bar; please fix your client./CN=sample.name"`
- Java `"CN=sample.name,OU=foo bar\\; please fix your client."`

PR #214, opened 2023-06-22
reduce error logging on connection resets

kaisecheng:
Fixed: #215

Change the log level to info for connection reset


edmocosta (changes requested):

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -222,9 +226,8 @@ public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cau
             private boolean silentException(final Throwable ex) {
                 if (ex instanceof IOException) {
                     final String message = ex.getMessage();
-                    if ("Connection reset by peer".equals(message)) {
-                        return true;
-                    }
+                    if (message != null)
+                        return silentErrs.stream().anyMatch(message::contains);
```
mashhurs, line 230: It seems to me we can simplify the `Array` since both elements have `Connection reset` message:
```suggestion
                    if (Objects.nonNull(message))
                        return message.contains(CONNECTION_RESET_ERROR_MSG);
```

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -188,6 +190,8 @@ private static final class DecoderAdapter extends ChannelInboundHandlerAdapter {
              */
             private final Logger logger;
 
+            private final List<String> silentErrs = Arrays.asList("Connection reset by peer", "Connection reset");
```
mashhurs, line 193:
```suggestion
            private final String CONNECTION_RESET_ERROR_MSG = "Connection reset";
```

src/main/java/org/logstash/tcp/InputLoop.java
mashhurs: if there is `StringUtils` available, using `nonBlank` or `isEmpty` would be also nice.
```
if (ex instanceof IOException == FALSE) {
  return false;
}

final String errorMessage = ex.getMessage();
return Objects.nonNull(errorMessage) ? errorMessage.contains(CONNECTION_RESET_ERROR_MSG): false;
```

mashhurs (commented):
Put my thoughts!

src/main/java/org/logstash/tcp/InputLoop.java
kaisecheng: You are right! It is just "Connection reset".

src/main/java/org/logstash/tcp/InputLoop.java
kaisecheng: updated

PR #220, opened 2024-04-23
Update Netty to 4.1.109

andsel:


kaisecheng (approved):

skumarp7:
Hi @roaksoax , 

I had raised an similar PR for the same version bump on Netty couple of weeks ago. Just curious to know if there was anything missing in that PR? I could see similar changes in this PR also except for the latest version of Netty.

jsvd:
@skumarp7 Nothing was missing other than 108 vs 109 and the explicit cve reference - which we tend to avoid, both could have been easily addressed. You had the CLA signed and everything.
Apologies for this, it was just that the person who was tasked with upgrading netty internally went for it and created the PRs across the three plugins and didn't notice the existing ones.

PR #222, opened 2024-07-03
Improve Log4j 2 example

vy:
- Switch from deprecated JSON Layout to its successor JSON Template Layout
- Use variable names in capitals (the recommended convention by Log4j)
- Explain configuration using AsciiDoc callouts
- Use ECS-compliant layout (which avoids the need to define a Logstash filter)

This work is delivered as a part of apache/logging-log4j2#2510.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -28,45 +28,46 @@ Like stdin and file inputs, each event is assumed to be one line of text.
 Can either accept connections from clients or connect to a server,
 depending on `mode`.
 
-===== Accepting log4j2 logs
+===== Accepting Log4j 2 logs
 
-Log4j2 can send JSON over a socket, and we can use that combined with our tcp
-input to accept the logs. 
+Log4j 2 can write ECS-compliant JSON-formatted log events to a TCP socket.
+We can combine with our TCP input to accept the logs from applications using Log4j 2.
 
-First, we need to configure your application to send logs in JSON over a
-socket. The following log4j2.xml accomplishes this task.
+First, we need to configure your application to write JSON-formatted logs to a TCP socket:
 
-Note, you will want to change the `host` and `port` settings in this
-configuration to match your needs.
+.Example `log4j2.xml` configuration for writing JSON-formatted logs to Logstash TCP input
+[source,xml]
+----
+<Configuration xmlns="https://logging.apache.org/xml/ns"
+               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+               xsi:schemaLocation="
+                       https://logging.apache.org/xml/ns
+                       https://logging.apache.org/xml/ns/log4j-config-2.xsd">
+  <Appenders>
+    <Socket name="SOCKET" host="localhost" port="12345"> <!--1-->
+      <JsonTemplateLayout <!--2-->
+          eventTemplateUri="classpath:EcsLayout.json" <!--3-->
+          nullEventDelimiterEnabled="true"/> <!--4-->
+    </Socket>
+  </Appenders>
+  <Loggers>
+    <Root level="INFO">
+      <AppenderRef ref="SOCKET"/>
+    </Root>
+  </Loggers>
+</Configuration>
+----
+<1> Using Socket Appender to write logs to a TCP socket – make sure to *change the `host` attribute* to match your setup
+<2> Using https://logging.apache.org/log4j/2.x/manual/json-template-layout.html[JSON Template Layout] to encode log events in JSON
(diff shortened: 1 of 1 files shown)


vy:
@jsvd, is there anything else I can provide further to get this merged and/or speed things up?

docs/index.asciidoc
```diff
@@ -28,31 +28,41 @@ Like stdin and file inputs, each event is assumed to be one line of text.
...
+    </Socket>
+  </Appenders>
+  <Loggers>
+    <Root level="INFO">
+      <AppenderRef ref="SOCKET"/>
+    </Root>
+  </Loggers>
+</Configuration>
+----
+<1> Using Socket Appender to write logs to a TCP socket – make sure to *change the `host` attribute* to match your setup
```
yaauie, line 60:
```suggestion
<1> Using Socket Appender to write logs to a TCP socket – make sure to *change the `host` and `port` attributes* to match your setup
```

docs/index.asciidoc
```diff
@@ -28,31 +28,41 @@ Like stdin and file inputs, each event is assumed to be one line of text.
...
+<2> Using https://logging.apache.org/log4j/2.x/manual/json-template-layout.html[JSON Template Layout] to encode log events in JSON
+<3> Using the ECS (Elastic Common Schema) layout bundled with JSON Template Layout
+<4> Configuring that written log events should be terminated with a null (i.e., `\0`) character
```
yaauie, line 63: This is surprising to me, and results in errors.

From the docs, this config will cause each event to be terminated with a two-byte newline-null sequence.

When the TCP input is configured with `codec => json` or `codec => json_lines`, it uses the `json_lines` codec, which splits the input on newline. This means that the post-`eventDelimiter` null byte will be the _first_ byte of each subsequent sequence passed to the codec, which results in json parse failures.

For example, if I start Logstash minimally:

~~~
bin/logstash -e 'input { tcp { codec => json port => 9887 } }'
~~~

And then send it json that is both newline and null delimited:

~~~
echo $'{"this":"that"}\n\0{"foo":"bar"}\n\0' | nc 127.0.0.1 9887
~~~ 

The null bytes cause `_jsonparsefailure`s:

> ~~~
> {
>       "@version" => "1",
>     "@timestamp" => 2025-02-18T17:37:51.852917Z,
>           "this" => "that"
> }
> {
>        "message" => "\u0000{\"foo\":\"bar\"}",
>           "tags" => [
>         [0] "_jsonparsefailure"
>     ],
>       "@version" => "1",
>     "@timestamp" => 2025-02-18T17:37:51.855102Z
> }
> {
>        "message" => "\u0000",
>           "tags" => [
>         [0] "_jsonparsefailure"
>     ],
>       "@version" => "1",
>     "@timestamp" => 2025-02-18T17:37:51.855359Z
> }
> ~~~

These issues do _not_ occur if we do not have the post-`eventDelimiter` null byte:

~~~
echo $'{"this":"that"}\n{"foo":"bar"}\n' | nc 127.0.0.1 9887
~~~

> ~~~
> {
>       "@version" => "1",
>     "@timestamp" => 2025-02-18T17:47:34.570363Z,
>           "this" => "that"
> }
> {
>       "@version" => "1",
>     "@timestamp" => 2025-02-18T17:47:34.570605Z,
>            "foo" => "bar"
> }
> ~~~

vy: You're right – I've confused it with the GELF Input Plugin, which requires null termination.

As a matter of fact, we have an IT for ELK:

* [`docker-maven-plugin` is used to bootstrap the ELK stack](/apache/logging-log4j2/blob/85c6c9b8cf18c1458e3abb5ff60bf233f4acf85b/log4j-layout-template-json-test/pom.xml#L195)
* [GELF Layout is configured with a null terminator](/apache/logging-log4j2/blob/85c6c9b8cf18c1458e3abb5ff60bf233f4acf85b/log4j-layout-template-json-test/src/test/java/org/apache/logging/log4j/layout/template/json/LogstashIT.java#L93)
* [ECS Layout isn't provided any delimiter, hence, falls back to the default, i.e., a newline](/apache/logging-log4j2/blob/85c6c9b8cf18c1458e3abb5ff60bf233f4acf85b/log4j-layout-template-json-test/src/test/java/org/apache/logging/log4j/layout/template/json/LogstashIT.java#L117)

Also, in apache/logging-log4j2@fef8af8e66981bf32933b2158cafeabc3305c2d0, fixed our [SOA page guiding users on ELK integration](https://logging.apache.org/log4j/2.x/soa.html#ELK).

PR #226, opened 2024-11-14
remove .travis.yml customization

jsvd:


donoghuc:
CI summary:
- for `ELASTIC_STACK_VERSION=8.previous` that will currently resolve to `8.15.4` which will be red due to https://github.com/elastic/logstash/issues/16694
- for `ELASTIC_STACK_VERSION=8.current` that will currently resolve to 8.16.0. That does not have the same bug as 8.15.4, but there appears to be an ipv6 address observed when we expect ipv4 (looking in to this still)
```
logstash-1  |   1) LogStash::Inputs::Tcp `ecs_compatibility => v8` should read plain with unicode
logstash-1  |      Failure/Error: expect(event.get(ecs_select[disabled: "[@metadata][ip_address]", v1: "[@metadata][input][tcp][source][ip]"])).to eq('127.0.0.1')
logstash-1  |      
logstash-1  |        expected: "127.0.0.1"
logstash-1  |             got: "0:0:0:0:0:0:0:1"
logstash-1  |      
logstash-1  |        (compared using ==)
logstash-1  |      # ./spec/inputs/tcp_spec.rb:93:in `block in <main>'
logstash-1  |      # ./spec/inputs/tcp_spec.rb:90:in `block in <main>'
logstash-1  |      # ./spec/inputs/tcp_spec.rb:87:in `block in times'
logstash-1  |      # ./spec/inputs/tcp_spec.rb:87:in `block in <main>'
logstash-1  |      # ./Rakefile:14:in `block in <main>'
```
- For `ELASTIC_STACK_VERSION=main` that currently resolves to  `9.0.0-SNAPSHOT` which has everything it needs merged into the main branch of logstash, but we are waiting for a DRA snapshot build
- For `ELASTIC_STACK_VERSION=8.future ` that currently resolves to `8.17.0-SNAPSHOT`. That is waiting on a backport pr https://github.com/elastic/logstash/pull/16705 and subesquent DRA snapshot build.

donoghuc:
Well, i went to reproduce the ipv4 vs ipv6 issue and I could not. The only failures left are the same encoding ones as are in flight. I will keep an eye on this to see if the issues come back. I see a couple places where there may be non deterministic behavior when resolving between v4 and v6 but I cant prove it. 👻

jsvd:
It did happen again 2 hours ago: https://app.travis-ci.com/github/logstash-plugins/logstash-input-tcp/jobs/628356013#L1116, so it's probably best to keep the validation for both ipv6 and ipv4

donoghuc:
Ok. Just deleted the revert commit for the work around in the test. I went on a goose chase last week trying to nail down where exactly that behavior was showing up but I was never able to actually put my finger on it. Unfortunately I didnt really take good notes on where all I went with it 🙃

jsvd:
The build is now failing only due to main and 8.x snapshots not having been built yet across the stack, so we can merge this.

donoghuc (approved):

PR #238, opened 2025-10-28
use PortManagementSupport mixin to reserve port in #register

yaauie:
Uses the `PortManagementSupport` mixin from https://github.com/logstash-plugins/logstash-mixin-port_management_support/pull/1 to reserve a port in `#register` so that the plugin will fail to start if the port is unavailable.

Resolves the TCP input's expression of https://github.com/elastic/logstash/issues/17703

### DIFF
--- a/lib/logstash/inputs/tcp.rb
+++ b/lib/logstash/inputs/tcp.rb
@@ -6,6 +6,7 @@ require "logstash/inputs/base"
 require "logstash/util/socket_peer"
 require "logstash-input-tcp_jars"
 require 'logstash/plugin_mixins/ecs_compatibility_support'
+require 'logstash/plugin_mixins/port_management_support'
 
 require "socket"
 require "openssl"
@@ -67,6 +68,8 @@ class LogStash::Inputs::Tcp < LogStash::Inputs::Base
 
   # ecs_compatibility option, provided by Logstash core or the support adapter.
   include LogStash::PluginMixins::ECSCompatibilitySupport(:disabled, :v1, :v8 => :v1)
+
+  include LogStash::PluginMixins::PortManagementSupport
 
   config_name "tcp"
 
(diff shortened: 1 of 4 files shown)


lib/logstash/inputs/tcp.rb
```diff
@@ -177,15 +180,20 @@ def register
     validate_ssl_config!
 
     if server?
-      @loop = InputLoop.new(@id, @host, @port, DecoderImpl.new(@codec, self), @tcp_keep_alive, java_ssl_context)
+      @port_reservation = port_management.reserve(addr: @host, port: @port) do |reserved_addr, reserved_port|
+        @loop = InputLoop.new(@id, reserved_addr, reserved_port, DecoderImpl.new(@codec, self), @tcp_keep_alive, java_ssl_context)
+      end
```
jsvd: I don't think we need to use blocks here since we're not wrapping behavior:

```suggestion
      port_management.reserve(port: @port) # if this succeeds, there is a reservation for the port
      @loop = InputLoop.new(@id, @host, @port, DecoderImpl.new(@codec, self), @tcp_keep_alive, java_ssl_context)
```

Also we should set the reservation scope for the port alone. Not sure if it's worth differentiating the addr, we can be conservative here and allow the port to be reserved regardless of the addr.

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -82,11 +87,26 @@ public InputLoop(final String id, final String host, final int port, final Decod
...
+            channel = serverBootstrap.bind(host, port).sync().channel();
+        }catch (final InterruptedException ex) {
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    public void waitUntilClosed() {
+        synchronized (this) {
+            if (channel == null) {
+                throw new IllegalStateException("not started");
```
jsvd:
```suggestion
                throw new IllegalStateException("Not started");
```

spec/inputs/tcp_spec.rb
```diff
@@ -21,19 +21,27 @@
...
+    with_bound_port(&:itself)
+  end
+
+  ##
+  # Yields block with a port that is unavailable
+  # @yieldparam port [Integer]
+  # @yieldreturn [Object]
+  # @return [Object]
+  def with_bound_port(port=0, &block)
+    server = TCPServer.new("::", port)
```
jsvd: tests used to focus on ipv4 and while I'm sure we also want to move tests to ipv6 not sure if this PR is the right place to do it.

jsvd (approved):
few tiny tweaks otherwise LGTM

jsvd (approved):

PR #250, opened 2026-04-14
fix: bind to port in `#register` to propagate bind failures

yaauie:
This PR supercedes #238:

| PR | `#register` | `#run` | Behaviour |
| --- | ------ | ------ | ------ |
| #238 | bind temporary TCP server | close and re-bind | Clients connecting between `#register` and `#run` get `TCP RST` |
| this | bind actual TCP server with auto-read off | turn auto-read on | Clients connecting after `#register` wait on the TCP `SO_BACKLOG` queue until `#run` |


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.0.8
+  - When configured to use a port that is already in use, the failure is now propagated to the pipeline
```
yaauie:
```suggestion
  - When configured to use a port that is already in use, the failure is now propagated to the pipeline [#250](https://github.com/logstash-plugins/logstash-input-tcp/pull/250)
```

lib/logstash/inputs/tcp.rb
```diff
@@ -504,4 +510,18 @@ def error_details(e, trace = false)
...
+  # https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/
+  def self._loop_close_proc(server)
+    proc do
+      begin
+        logger.trace("ObjectSpace::closing(#{server})")
+        server.close
+        logger.trace("ObjectSpace::closed(#{server})")
+      rescue Exception => e
+        logger.error("Failed to close #{server}: #{e.message}")
+      end
```
yaauie:
```suggestion
      Thread.new do # don't block the GC thread
        begin
          logger.trace("ObjectSpace::closing(#{server})")
          server.close
          logger.trace("ObjectSpace::closed(#{server})")
        rescue Exception => e
          logger.error("Failed to close #{server}: #{e.message}")
        end
      end
```

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -75,17 +75,26 @@ public InputLoop(final String id, final String host, final int port, final Decod
...
             .channel(NioServerSocketChannel.class)
             .option(ChannelOption.SO_BACKLOG, 1024)
+            .option(ChannelOption.AUTO_READ, false) // do not auto-read until plugin has a queue
             .childOption(ChannelOption.SO_KEEPALIVE, keepAlive)
             .childHandler(new InputLoop.InputHandler(decoder, sslContext));
+
+        try {
+            serverChannel = serverBootstrap.bind(host, port).sync().channel();
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
donoghuc, line 90:
```suggestion
            boss.shutdownGracefully();
            worker.shutdownGracefully();
            throw new RuntimeException(e);
```
Should we attempt to clean up these resources?

donoghuc (changes requested):
Manual testing validates this approach works well! 

I had one suggestion that i think could help.

donoghuc (approved):
LGTM!

PR #286, opened 2026-09-01
Aggregate HA proxy line before parsing it

### LINKED ISSUE #285: Proxy protocol erroneously parsed
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
4. How was the Logstash Plugin installed

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc).
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):

**Description of the problem including expected versus actual behavior**:

**Steps to reproduce**:

The failure happened in a test https://github.com/logstash-plugins/logstash-input-tcp/actions/runs/31784048596/job/94715876619#step:4:272

The problem is the proxy protocol parsing at https://github.com/logstash-plugins/logstash-input-tcp/blob/f57856665de24901a2be55d31f680fa10368260b/lib/logstash/inputs/tcp/decoder_impl.rb#L39 which expect to have a full line with `\r\n`  present in the received buffer. The buffer comes from the network layer and this isn't any guarantee that the expected line is present.
For example, bringing from test https://github.com/logstash-plugins/logstash-input-tcp/blob/f57856665de24901a2be55d31f680fa10368260b/spec/inputs/tcp_spec.rb#L140

```
PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r\n
```

can be spitted in  2 buffers:
```
PROXY TCP4 1.2.3.4 5.6.7.8 1234
```

and

```
 5678\r\n
```

and the `received.split("\r\n", 2)` would return just one element array, like
```
["PROXY TCP4 1.2.3.4 5.6.7.8 123"]
```
this means that `filtered` is nil and is passed down to the codec, which sends to the BufferedTokenizerExt , where `data.toString()` throws an NPE.

In the case of the test, the NPE is logged, but doesn't stop client threads to wait indefinitely on the `queue.pop` at https://github.com/logstash-plugins/logstash-input-tcp/blob/f57856665de24901a2be55d31f680fa10368260b/spec/inputs/tcp_spec.rb#L149 blocking the test and CI for ever.

**Provide logs (if relevant)**:

andsel:
## Release notes
Fix for processing the HA Proxy protocol, considering the line can be splitted in multiple buffers.

## What does this PR do?
Adds an aggregator to collect all the bytes up to the first newline before move on to the decoding part done by DecoderImpl class. This message decoder is added only if proxy setting is enabled.

## Why is it important/What is the impact to the user?
Avoids error that could rise if the TCP stack decides to break the HA Proxy v1 headline before the newline terminator in multiple byte arrays.

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~
- [x] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [ ]

## Related issues

-  Closes #285

### DIFF
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -42,7 +42,8 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
   end
 
   def queue_pop_with_timeout(seconds, queue)
-    deadline = Time.now + seconds
+    start = Time.now
+    deadline = start + seconds
     item = nil
     until item || Time.now >= deadline
       begin
@@ -50,6 +51,9 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
       rescue ThreadError
         sleep 0.1
       end
+    end
+    if item == nil
+      raise "Elapsed #{Time.now - start} seconds before pop a value"
     end
     return item
   end
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -41,6 +41,19 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
     server.close
   end
 
+  def queue_pop_with_timeout(seconds, queue)
+    deadline = Time.now + seconds
+    item = nil
+    until item || Time.now >= deadline
+      begin
+        item = queue.pop(true)   # raises ThreadError if empty
+      rescue ThreadError
+        sleep 0.1
+      end
+    end
+    return item
+  end
+
   let(:port) { find_available_port("127.0.0.1") }
 
   context "codec (PR #1372)" do
@@ -139,15 +152,23 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
         socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
         socket.write("PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r")
         socket.flush
-        socket.write("\n")
-        event_count.times do |i|
-          # unicode smiley for testing unicode support!
-          socket.puts("#{i} ☹")
-          socket.flush
+        # proceed with the rest of data writing in a separate thread
+        # so that crates some interleaving in the buffers seen on network layer
+        Thread.new do
+          sleep(1)
+          begin
+            socket.write("\n")
+            event_count.times do |i|
+              # unicode smiley for testing unicode support!
+              socket.puts("#{i} ☹")
+              socket.flush
+            end
+            socket.close
+          rescue => e
+            puts "Error while writing to the socket #{e.backtrace}"
+          end
         end
-        socket.close
-
-        event_count.times.collect {queue.pop}
+        event_count.times.collect { queue_pop_with_timeout(5, queue) }
       end
 
       expect(events.length).to eq(event_count)
@@ -675,6 +696,7 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
       context "when ssl_enabled is true" do
         let(:input) { subject }
         let(:queue) { Queue.new }
+        # let(:queue) { Thread::Queue.new }
         before(:each) do
           allow_any_instance_of(described_class).to receive(:ecs_compatibility).and_return(ecs_compatibility) if defined?(ecs_compatibility)
           subject.register
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -700,7 +700,6 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
       context "when ssl_enabled is true" do
         let(:input) { subject }
         let(:queue) { Queue.new }
-        # let(:queue) { Thread::Queue.new }
         before(:each) do
           allow_any_instance_of(described_class).to receive(:ecs_compatibility).and_return(ecs_compatibility) if defined?(ecs_compatibility)
           subject.register
--- a/spec/inputs/tcp_spec.rb
+++ b/spec/inputs/tcp_spec.rb
@@ -135,10 +135,11 @@ describe LogStash::Inputs::Tcp, :ecs_compatibility_support do
         }
       CONFIG
 
-      events = input(conf) do |pipeline, queue|
+      events = input(conf) do |_, queue|
         socket = Stud::try(5.times) { TCPSocket.new("127.0.0.1", port) }
-        socket.puts("PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r");
+        socket.write("PROXY TCP4 1.2.3.4 5.6.7.8 1234 5678\r")
         socket.flush
+        socket.write("\n")
         event_count.times do |i|
           # unicode smiley for testing unicode support!
           socket.puts("#{i} ☹")
--- a/src/main/java/org/logstash/tcp/InputLoop.java
+++ b/src/main/java/org/logstash/tcp/InputLoop.java
@@ -69,7 +69,7 @@ public final class InputLoop implements Runnable, Closeable {
      * @param keepAlive set to true to instruct the socket to issue TCP keep alive
      */
     public InputLoop(final String id, final String host, final int port, final Decoder decoder, final boolean keepAlive,
-                     final SslContext sslContext) {
+                     final SslContext sslContext, final boolean proxy) {
         this.sslContext = sslContext;
         this.host = host;
         this.port = port;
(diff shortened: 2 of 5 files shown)


andsel:
**Note for reviewer**

- the commit c361409b60890b4bfa37c4e5768f9314a29b6505 clearly made evident the problem with partial processing of HA proxy header, when a non complete line is parsed by `DecoderImpl` it return a `nil` that crashes the processing making the pop block on something that is never sent. The commit add also a timeout to the Queue.pop operation to avoid waiting indefinitely. This is 🔴 CI
- the commit ea983d52a0a13f0fec29b18370337c867c6849fb enable the ProxyLineAgrgeagtor plus another fix needed to correctly identify the sequence of `\r\n` in the ByteBuf. This is 🟢 CI

spec/inputs/tcp_spec.rb
```diff
@@ -674,6 +700,7 @@ def with_bound_port(host:"::", port:0, &block)
       context "when ssl_enabled is true" do
         let(:input) { subject }
         let(:queue) { Queue.new }
+        # let(:queue) { Thread::Queue.new }
```
donoghuc:
```suggestion
```

src/main/java/org/logstash/tcp/InputLoop.java
```diff
@@ -69,7 +69,7 @@ public final class InputLoop implements Runnable, Closeable {
      * @param keepAlive set to true to instruct the socket to issue TCP keep alive
```
donoghuc, line 69:
```suggestion
     * @param keepAlive set to true to instruct the socket to issue TCP keep alive
     * @param sslContext SSL configuration, or null when TLS is disabled
     * @param proxy set to true to aggregate the HAProxy v1 header line before decoding
```

src/main/java/org/logstash/tcp/ProxyLineAggregator.java
```diff
@@ -0,0 +1,69 @@
...
+
+
+/**
+ * Accumulate bytes until the HAProxy format v1 headline is present in the buffer.
+ * The line has format "PROXY....\r\n", this aggregator reject the buffer until it has that format,
+ * once reached passthrough every buffer as it is.
+ * This is needed because Ruby class DecoderImpl expect to have the full line when processing HAProxy protocol, and
+ * doesn't work with fragments.
+ * */
+public class ProxyLineAggregator extends ByteToMessageDecoder {
```
donoghuc, line 18: I feel the least strongly about this, but an LLM suggested the following:

```diff
diff --git a/src/main/java/org/logstash/tcp/ProxyLineAggregator.java b/src/main/java/org/logstash/tcp/ProxyLineAggregator.java
index 918a1ce..c5b33b9 100644
--- a/src/main/java/org/logstash/tcp/ProxyLineAggregator.java
+++ b/src/main/java/org/logstash/tcp/ProxyLineAggregator.java
@@ -11,46 +11,30 @@ import java.util.List;
 
 /**
  * Accumulate bytes until the HAProxy format v1 headline is present in the buffer.
- * The line has format "PROXY....\r\n", this aggregator reject the buffer until it has that format,
- * once reached passthrough every buffer as it is.
- * This is needed because Ruby class DecoderImpl expect to have the full line when processing HAProxy protocol, and
- * doesn't work with fragments.
+ * The line has format "PROXY....\r\n"; this aggregator holds back the buffer until that full
+ * line is available, then passes it through and removes itself from the pipeline.
+ * This is needed because Ruby class DecoderImpl expects the full line when processing the HAProxy
+ * protocol and doesn't work with fragments.
  * */
 public class ProxyLineAggregator extends ByteToMessageDecoder {
 
     private static final byte[] PROXY_PREFIX = "PROXY".getBytes(StandardCharsets.US_ASCII);
     public static final int PROXY_LENGTH = PROXY_PREFIX.length;
 
-    enum DecoderState {READ_PROXY, COMPLETED}
-
-    private DecoderState state;
-
-    public ProxyLineAggregator() {
-        this.state = DecoderState.READ_PROXY;
-    }
-
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
-        switch (state) {
-            case READ_PROXY:
-                if (buffer.readableBytes() < PROXY_LENGTH) {
-                    return;
-                }
-                if (!startsWithProxy(buffer)) {
-                    state = DecoderState.COMPLETED;
-                    out.add(buffer.readRetainedSlice(buffer.readableBytes()));
-                    return;
-                }
-                if (!containsCrlf(buffer)) {
-                    return;
-                }
-                state = DecoderState.COMPLETED;
-                out.add(buffer.readRetainedSlice(buffer.readableBytes()));
-                break;
-            case COMPLETED:
-                out.add(buffer.readRetainedSlice(buffer.readableBytes()));
-                break;
+        // Wait until we can decide: enough bytes to match the prefix, and if it is a PROXY
+        // line, the terminating \r\n must be present.
+        if (buffer.readableBytes() < PROXY_LENGTH) {
+            return;
+        }
+        if (startsWithProxy(buffer) && !containsCrlf(buffer)) {
+            return;
         }
+        // Full PROXY line, or non-PROXY data: pass everything through and drop this handler
+        // so subsequent reads skip the aggregator entirely.
+        out.add(buffer.readRetainedSlice(buffer.readableBytes()));
+        ctx.pipeline().remove(this);
     }
 
     private static boolean containsCrlf(ByteBuf buffer) {
```

Piece by piece:

- **Javadoc rewrite:** Fixes the grammar ("reject" → "holds back", "expect" →
  "expects") and, importantly, documents the new behavior: the handler now
  *removes itself* once done, instead of lingering forever in `COMPLETED`.

- **Deleted `enum DecoderState`, the `state` field, and the constructor:** These
  existed only to remember "have I already resolved the header?" Once the handler
  removes itself from the pipeline, there is no second invocation to remember
  anything for, so the whole state machine becomes dead weight.

- **New `decode` body — three steps:**
  1. `if (buffer.readableBytes() < PROXY_LENGTH) return;` — not enough bytes yet
     to even compare against the 5-byte `PROXY` prefix, so wait for more.
     `ByteToMessageDecoder` re-invokes `decode` as more bytes arrive, accumulating
     them in its internal cumulation buffer.
  2. `if (startsWithProxy(buffer) && !containsCrlf(buffer)) return;` — it *looks*
     like a PROXY line but the `\r\n` terminator hasn't arrived yet, so keep
     waiting. This is exactly the bug from issue #285: the header split across
     TCP buffers. If it does **not** start with `PROXY`, this condition is false
     and we fall through immediately (non-proxy traffic passes straight through).
  3. `out.add(buffer.readRetainedSlice(buffer.readableBytes()));` then
     `ctx.pipeline().remove(this);` — forward everything currently buffered
     downstream to the decoder, then unhook the aggregator. `readRetainedSlice`
     hands off a reference-counted slice without copying the bytes; the emitted
     slice is added to `out` *before* removal, so it is still delivered.

- **Behavioral equivalence:** The old `COMPLETED` branch just passed buffers
  through unchanged. Removing the handler achieves the same downstream result for
  all subsequent bytes, but with zero per-read overhead (no switch, no slice copy,
  no re-entry) after the header is resolved. All four existing JUnit cases —
  partial prefix, prefix-without-CRLF, non-proxy passthrough, complete line —
  still pass unchanged.

Note: `containsCrlf` and `startsWithProxy` (below this hunk) are untouched.

donoghuc (changes requested):
Take or leave some of these (especially the LLM one), i'll put a request change on this until we decide. But overall this is cool to see our tests catching this kind of bug!

src/main/java/org/logstash/tcp/ProxyLineAggregator.java
andsel: Seems a fair suggestion and removing things that aren't functionally useful is an housekeeping task. 
I've accepted it.

donoghuc (approved):
lgtm. Looks like we will want a release. just need a `pr_prepare` comment!

andsel:
/pr_prepare release:patch
-  Fix for processing the HA Proxy protocol, considering the line can be split in multiple buffers.

